PR #11, opened 2017-10-20
RFC: oui-switch

neolitec:

### DIFF
--- a/packages/oui-switch/src/switch.component.js
+++ b/packages/oui-switch/src/switch.component.js
@@ -1,0 +1,23 @@
+import controller from './switch.controller'
+
+export default {
+  template: `<label class="oui-switch">
+      <input type="checkbox"
+        name="{{ $ctrl.name }}"
+        class="oui-switch__checkbox"
+        ng-disabled="$ctrl.disabled"
+        ng-model="$ctrl.value"
+        ng-change="$ctrl.onChange({ $event: { value: $ctrl.value } })">
+      <div class="oui-switch__slider"></div>
+      <i class="oui-icon oui-icon-success" aria-hidden="true"></i>
+      <i class="oui-icon oui-icon-error" aria-hidden="true"></i>
+    </label>`,
+  controller,
+  bindings: {
+    disabled: '<?',
+    name: '@?',
+    value: '<',
+    onChange: '&?'
+  },
+  transclude: true
+}
--- a/packages/oui-switch/src/switch.controller.js
+++ b/packages/oui-switch/src/switch.controller.js
@@ -1,0 +1,18 @@
+import { addBooleanParameter } from '@oui-angular/common/component-utils'
+
+export default class {
+  constructor ($scope, $attrs, $transclude) {
+    'ngInject'
+
+    this.$scope = $scope
+    this.$attrs = $attrs
+  }
+
+  $onInit () {
+    addBooleanParameter(this, 'disabled')
+
+    if (!this.name) {
+      this.name = `oui-switch-${this.$scope.$id}`
+    }
+  }
+}
--- a/packages/common/component-utils.js
+++ b/packages/common/component-utils.js
@@ -1,0 +1,11 @@
+import _ from 'lodash'
+
+export function addBooleanParameter (controller, parameterName) {
+  if (_.has(controller.$attrs, 'disabled') && _.isEmpty(controller.$attrs.disabled)) {
+    controller.disabled = true
+  }
+}
+
+export default {
+  addBooleanParameter
+}
--- a/packages/common/component-utils.js
+++ b/packages/common/component-utils.js
@@ -1,7 +1,7 @@
 import _ from 'lodash'
 
 export function addBooleanParameter (controller, parameterName) {
-  if (_.has(controller.$attrs, 'disabled') && _.isEmpty(controller.$attrs.disabled)) {
+  if (_.has(controller.$attrs, parameterName) && _.isEmpty(controller.$attrs[parameterName])) {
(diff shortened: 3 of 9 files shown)


packages/oui-switch/src/switch.controller.js
```diff
@@ -0,0 +1,18 @@
...
+    'ngInject'
+
+    this.$scope = $scope
+    this.$attrs = $attrs
+  }
+
+  $onInit () {
+    addBooleanParameter(this, 'disabled')
+
+    if (!this.name) {
```
Mboulianne: maybe addStringParameter with a default value? (like addBooleanParameter)

packages/oui-switch/src/switch.component.js
```diff
@@ -0,0 +1,23 @@
+import controller from './switch.controller'
+
+export default {
+  template: `<label class="oui-switch">
```
jleveugle: What are the usecases of a switch compared to a checkbox ?

Can `onChange` event be async ?

packages/oui-switch/src/switch.component.js
```diff
@@ -0,0 +1,23 @@
...
+      <input type="checkbox"
+        name="{{ $ctrl.name }}"
+        class="oui-switch__checkbox"
+        ng-disabled="$ctrl.disabled"
+        ng-model="$ctrl.value"
+        ng-change="$ctrl.onChange({ $event: { value: $ctrl.value } })">
+      <div class="oui-switch__slider"></div>
+      <i class="oui-icon oui-icon-success" aria-hidden="true"></i>
+      <i class="oui-icon oui-icon-error" aria-hidden="true"></i>
+    </label>`,
```
jleveugle: How can i pass a associate text to my component ? We should hide a text in a label to have a textual alternative to the checkbox :)

Actually, the component won't be accessible :(

jleveugle: We should define the documentation and how to use the component before coding it imo :)

packages/oui-switch/src/switch.controller.js
```diff
@@ -0,0 +1,18 @@
+import { addBooleanParameter } from '@oui-angular/common/component-utils'
+
+export default class {
+  constructor ($scope, $attrs, $transclude) {
+    'ngInject'
+
+    this.$scope = $scope
```
jleveugle: What do you think about avoiding using $scope ?

I hope one day, we will just do a breaking release and pass all our component to ng4+ :)

jleveugle: We should have a look to : https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#controlleras-controller-syntax

neolitec: It's just an easy way to create a unique id base on the $id property of scope here.
Thank you for the doc I discover now all these rules (I'm kidding), unfortunately, avoiding using $scope in a huge components library ... and directives (yes, sorry) library in angular < 2 is a pure utopia.

packages/oui-switch/src/switch.component.js
neolitec: No, but it could be, just specify in which use case here please (even if I can guess it).

packages/oui-switch/src/switch.component.js
neolitec: Ok for the accessibility, it will have to be done.
The documentation hasn't been done yet
The aim of this RFC, I've forgotten to explain that, is also to discuss about the one-way and the two-way data binding we will probably adopt in the future of ovh-ui-angular.

packages/oui-switch/src/switch.controller.js
neolitec: Could be nicer indeed!

packages/oui-switch/src/switch.component.js
Mboulianne: I strongly disagree with doc before code.  If you want down to earth API / interface it's better to implement things based on a real life needs and not on some sort of documentation.

Remember, this is a component.  We can introduce change easily if we choose (or need) to.  I think the component in it's current form encompass what we need for step 1 => A dropdown for our action menu.  I'm not saying we should forget about the future, however, we need to stay away from over-design.

packages/oui-switch/src/switch.controller.js
Mboulianne: I strongly disagree with doc before code. If you want down to earth API / interface it's better to implement things based on a real life needs and not on some sort of documentation.

Remember, this is a component. We can introduce change easily if we choose (or need) to. I think the component in it's current form encompass what we need for step 1 => A dropdown for our action menu. I'm not saying we should forget about the future, however, we need to stay away from over-design.

packages/oui-switch/src/switch.component.js
Mboulianne: What are the usecases of a switch compared to a checkbox ? 
We often use switches if the action is immediate (1 click = 1 post / put)

jleveugle: Yes @Mboulianne , it's the answer i would like to read :p

Switches -> immediate so async event won't be possible i think or we need to manage loader in switch

Mboulianne: yeah but why not deal with this specific issue when we encounter it?

neolitec: Or loader instead of switch...
I also think about the case of error: the use turns on the switch, the post request returns an error so that the switch should be turned back off.
The question is: is this logic must be supported by the component itself or by the controller (outside of the component)?

PR #13, opened 2017-11-15
feat(oui-button): add oui-button component

AxelPeter:
http://angular_feature_oui_button.ui-kit.ovh/#!/oui-angular/button


packages/oui-button/src/button.controller.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  // Check presence of attributes for modifiers
+  checkModifiers () {
+    this.primary = this.$attrs.hasOwnProperty('primary')
+    this.secondary = this.$attrs.hasOwnProperty('secondary')
+    this.link = this.$attrs.hasOwnProperty('link')
+
+    // Warning for multiple declaration
+    this.checkAttrs([
```
jimmyfortinx: Instead to pass two time attributes names maybe we could just call it with this signature `checkAttrs('primary', 'secondary', 'link')`. With this signature we could also do the `this.$attrs.hasOwnProperty(attributeName)` in this method.

Also, maybe `checkAttrs` is not best name because it doesn't specify what is the check.

So, could it be better to do this method:

```javascript
createOrAttrs (...attributes) {
  attributes.forEach((attributeName) => {
     this[attributeName] = this.$attrs.hasOwnProperty(attributeName)
  });

  if (_.sum(_.values(_.pick(this, attributes))) > 1) {
     var formattedAttributes = attributes.map((attributeName) => _.kebabCase(attributeName)).join('|')

     this.$log.warn(`Unauthorized multiple declaration of attributes (${formattedAttributes})`)
  }
}
```

This method should also be extracted in a service since not just the button will have this kind of attributes.

packages/oui-button/src/button.controller.js
```diff
@@ -0,0 +1,69 @@
...
+    ], 'primary|secondary|link')
+
+    // Default modifier style
+    if (!this.primary && !this.secondary && !this.link) {
+      this.secondary = true
+    }
+  }
+
+  // Check presence of attributes for steps
+  checkSteps () {
```
jimmyfortinx: Could be simplified by the `this.createOrAttrs("previousStep", "nextStep")` method.

packages/oui-button/src/button.controller.js
```diff
@@ -0,0 +1,69 @@
...
+
+    // Warning for multiple declaration
+    this.checkAttrs([
+      this.nextStep,
+      this.previousStep
+    ], 'next-step|previous-step')
+  }
+
+  // Check type for button (default is 'button')
+  checkType () {
```
jimmyfortinx: Proposition:

You can keep the `checkType` method but you could create a new method, at the same place than `createOrAttrs`, to simplify those kind of scenarios.
 
```javascript
createAttributeWithOrValues(attributeName, values, defaultValue) {
  if (_.indexOf(values, this[attributeName]) !== -1) {
     this[attributeName] = defaultValue
  }
}
```

packages/oui-button/src/button.html
```diff
@@ -0,0 +1,20 @@
+<button
+  arial-label="{{$ctrl.text}}"
+  class="oui-button"
+  ng-attr-id="{{$ctrl.id}}"
```
jimmyfortinx: one time binding ?

packages/oui-button/src/button.html
```diff
@@ -0,0 +1,20 @@
...
+  ng-attr-name="{{$ctrl.name}}"
```
jimmyfortinx: one time binding ?

packages/oui-button/src/button.html
```diff
@@ -0,0 +1,20 @@
...
+  ng-attr-type="{{$ctrl.type}}"
```
jimmyfortinx: one time binding ?

packages/oui-button/src/button.controller.js
```diff
@@ -0,0 +1,69 @@
...
+export default class {
+  constructor ($attrs, $element, $log) {
+    'ngInject'
+
+    this.$attrs = $attrs
+    this.$element = $element
+    this.$log = $log
+  }
+
+  $onInit () {
```
jimmyfortinx: Could we support also this syntax for the `disabled` attribute ? `<oui-button disabled></oui-button>`

I think it can be done with this:

```javascript
if(this.$attrs.hasOwnProperty('disabled') && this.$attrs.disabled === "") {
  this.disabled = true
}
```

packages/oui-button/src/button.html
AxelPeter: Yes

packages/oui-button/src/button.html
AxelPeter: Yes

packages/oui-button/src/button.html
AxelPeter: Yes :)

packages/oui-button/src/button.controller.js
AxelPeter: Yes :)

packages/oui-button/src/button.controller.js
frenautvh: it looks too complicated to me, maybe something simpler like :
```
if (this.primary + this.secondary + this.link === 1) {
  // good
} else {
 // bad
}
```

frenautvh: my bad you did the same thin with sum

AxelPeter: Yes, good idea. But we should add this on a service. Can be useful for the others components.

packages/oui-button/src/button.controller.js
neolitec: I have a helper for that.

packages/oui-button/src/button.component.js
```diff
@@ -0,0 +1,17 @@
+import template from './button.html'
+import controller from './button.controller'
+
+export default {
+  bindings: {
+    text: '@',
+    id: '@?',
+    name: '@?',
+    type: '@?',
```
neolitec: Let me know if it's excessive:
If we create a binding for this attribute, you let the user thinking that the notation `type="{{type}}"` could make it dynamic. However this is not the case here: `type` will never be re-parsed internally if it changes (we assume that's right, the type of a button is almost never updated).
1) We could just rely on $attrs instead of bindings if the value is not dynamic.
2) The documentation should be more clear IMO if you omit "@?" (this point is valid only if the (1) is valid).

packages/oui-button/webpack.config.babel.js
```diff
@@ -0,0 +1,7 @@
+import merge from 'webpack-merge'
```
neolitec: Remove this file.

packages/oui-button/package.json
```diff
@@ -0,0 +1,31 @@
+{
```
neolitec: Remove this file.

packages/oui-button/protractor.conf.js
```diff
@@ -0,0 +1,8 @@
+import fs from 'fs'
```
neolitec: I don't think we will use Protractor for now. So you can remove it.

packages/oui-button/karma.conf.js
```diff
@@ -0,0 +1,13 @@
+import { configure as configureKarma } from '../../build/karma.single.conf'
```
neolitec: Remove this file.

packages/oui-button/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+## API
+
+| Attribute     | Type     | Binding | Values              | Default | Description                      |
+| ----          | ----     | ----    | ----                | ----    | ----                             |
+| on-click      | function | &?      |                     |         | click handler                    |
+| text          | string   | @       |                     |         | button text                      |
+| id            | string   | @?      |                     |         | id attribute of the button       |
+| name          | string   | @?      |                     |         | name attribute of the button     |
+| type          | string   | @?      | submit,button,reset | button  | click handler                    |
```
neolitec: This is not the click handler.

packages/oui-angular/package.json
```diff
@@ -10,6 +10,7 @@
     "karma:watch": "cross-env NODE_ENV=test babel-node ./node_modules/karma/bin/karma start ../../build/karma.global.conf.js --watch"
   },
   "dependencies": {
+    "@oui-angular/button": "^1.0.0",
```
neolitec: Dependencies not used anymore. (This file neither)

antleblanc:
What's about a loading state? (containing `<oui-loader>`)

packages/oui-angular/package.json
AxelPeter: Ok, removed

packages/oui-button/README.md
AxelPeter: Corrected (it was in the file in sprint1)

packages/oui-button/karma.conf.js
AxelPeter: done

packages/oui-button/package.json
AxelPeter: done

packages/oui-button/protractor.conf.js
AxelPeter: done

packages/oui-button/webpack.config.babel.js
AxelPeter: done

packages/oui-button/src/button.component.js
AxelPeter: If I understand, you would remove `type: '@?'` and add 'this.type = this.$attrs['type']` on $onInit ?

IMO, I think we should keep it simple.
It's clearer for the dev (listed in bindings) and used in an Angular way.

jimmyfortinx: I think @neolitec is right, we should not let the user think that attribut is a real binding when we don't support when value is updated. Anyway, `type` is not the only attribute that should be documented outside bindings, `primary|secondary|link` for example. Also, we are giving a `README.md` file fully documented about that, I think we can do it.

AxelPeter: Seen together, we'll add it on the doc that it's a one-time binding :)

packages/oui-button/README.md
```diff
@@ -0,0 +1,76 @@
...
+  <oui-button primary next-step text="Next" on-click="next()"></oui-button>
+</div>
+
+<div>
+  <oui-button previous-step text="Previous" disabled="true" on-click="previous()"></oui-button>
+  <oui-button primary next-step text="Next" disabled="true" on-click="next()"></oui-button>
+</div>
+```
+
+### Accessibility
```
jimmyfortinx: h3 ? ça ne devrait pas plutot être au même niveau qu'API et Examples ?

packages/oui-button/src/button.html
```diff
@@ -0,0 +1,20 @@
...
+    'oui-button_secondary': $ctrl.variant === 'secondary',
+    'oui-button_link': $ctrl.variant === 'link',
+    'oui-button_icon-left': $ctrl.variantNav === 'previous',
+    'oui-button_icon-right': $ctrl.variantNav === 'next'
+  }"
+  ng-click="$ctrl.ngClick()"
+  ng-disabled="$ctrl.disabled"
+>
+  <span aria-hidden="true" class="oui-icon oui-icon_circle oui-icon-chevron-left" ng-if="$ctrl.variantNav === 'previous'"></span>
+  {{$ctrl.text}}
```
jimmyfortinx: [To talk at our next daily] Normally, at OVH we choose to prefer `<span data-ng-bind="$ctrl.text"></span>`. Is this requirement still needed or we can use `{{}}` syntax directly in the html ?

packages/oui-button/src/button.controller.js
```diff
@@ -0,0 +1,26 @@
...
+    if (angular.isUndefined(this.type)) {
+      this.type = 'button'
+    }
+
+    if (angular.isUndefined(this.variant)) {
+      this.variant = 'secondary'
+    }
+
+    this.disabled = angular.isDefined(this.$attrs.disabled) && angular.isUndefined(this.disabled)
```
jimmyfortinx: If the user does `<oui-button disabled="$ctrl.test"></oui-button>`, where `$ctrl.test` is `true`, are you sure it will not overwrite the parent value with `false`?

packages/oui-button/README.md
```diff
@@ -0,0 +1,73 @@
...
+  <oui-button variant="primary" variant-nav="next" text="Next" on-click="next()"></oui-button>
+</div>
+
+<div>
+  <oui-button variant-nav="previous-step" text="Previous" on-click="previous()" disabled></oui-button>
+  <oui-button variant="primary" variant-nav="next" text="Next" on-click="next()" disabled></oui-button>
+</div>
+```
+
+### Accessibility
```
jimmyfortinx: I think it should be at the same level than `Examples` and `API`, not under `Examples`.

packages/oui-button/src/button.html
```diff
@@ -0,0 +1,20 @@
+<button
+  arial-label="{{$ctrl.text}}"
+  class="oui-button"
+  ng-attr-id="{{::$ctrl.id}}"
+  ng-attr-name="{{::$ctrl.name}}"
+  ng-attr-type="{{::$ctrl.type}}"
+  ng-class="{
```
tremblaymath: You could add one time binding, I don't think these values will change after first digest
 ng-class=":: {...

packages/oui-button/src/button.html
```diff
@@ -0,0 +1,20 @@
+<button
+  arial-label="{{$ctrl.text}}"
```
tremblaymath: I don't think aria-label will change after first digest, one time binding

packages/oui-button/src/button.html
```diff
@@ -0,0 +1,20 @@
...
+  ng-attr-name="{{::$ctrl.name}}"
+  ng-attr-type="{{::$ctrl.type}}"
+  ng-class="{
+    'oui-button_primary': $ctrl.variant === 'primary',
+    'oui-button_secondary': $ctrl.variant === 'secondary',
+    'oui-button_link': $ctrl.variant === 'link',
+    'oui-button_icon-left': $ctrl.variantNav === 'previous',
+    'oui-button_icon-right': $ctrl.variantNav === 'next'
+  }"
+  ng-click="$ctrl.ngClick()"
```
tremblaymath: Using ng-click for a submit button should probably trigger a warning. We should force to use ng-submit on the form, it's a common problem. It could be a todo for the futur

packages/oui-button/README.md
```diff
@@ -0,0 +1,73 @@
...
+## API
+
+| Attribute     | Type     | Binding | Values                 | Default   | Description                      |
+| ----          | ----     | ----    | ----                   | ----      | ----                             |
+| on-click      | function | &?      |                        |           | click handler                    |
+| text          | string   | @       |                        |           | button text                      |
+| id            | string   | @?      |                        |           | id attribute of the button       |
+| name          | string   | @?      |                        |           | name attribute of the button     |
+| type          | string   | @?      | submit,button,reset    | button    | type attribute of the button     |
+| variant       | string   | @?      | primary,secondary,link | secondary | modifier for button              |
```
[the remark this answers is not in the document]
AxelPeter: Yes, i'll add it

packages/oui-button/src/button.html
```diff
@@ -0,0 +1,20 @@
+<button
+  arial-label="{{$ctrl.text}}"
+  class="oui-button"
+  ng-attr-id="{{::$ctrl.id}}"
```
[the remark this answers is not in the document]
AxelPeter: I'll add it, it's still a good thing

packages/oui-button/src/button.controller.js
AxelPeter: Yes, i think so... I'll put back in a `if`.

packages/oui-button/src/button.html
AxelPeter: Yes, we could do that with the text too

packages/oui-button/src/button.html
AxelPeter: Yes, done

packages/oui-button/src/button.html
AxelPeter: I don't know. The common problem is to use `ng-click` to submit a form. 
This thing must be verified with the PRs.

We can always have a case where we want to have a `ng-submit` and an action on the `ng-click` that is not a form submission.

packages/oui-button/README.md
AxelPeter: This part will be removed, because it'll have its own attribute.

packages/oui-button/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+<div>
+  <oui-button previous-step text="Previous" disabled="true" on-click="previous()"></oui-button>
+  <oui-button primary next-step text="Next" disabled="true" on-click="next()"></oui-button>
+</div>
+```
+
+### Accessibility
+
+`text` is used as an `aria-label`.
```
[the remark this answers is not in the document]
AxelPeter: I add another attribute `aria-label` for this case. 
It'll be removed from the root component like `id` and name`.

packages/oui-button/README.md
AxelPeter: I added an example to this part. I think it can stay in the examples area.
I move up the API part next to Usage.

packages/oui-button/src/index.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      let scope = $rootScope.$new()
+
+      $compile(element)(scope)
+      scope.$digest()
+
+      expect(element.attr('aria-label')).toBe(undefined)
+      expect(element.find('button').eq(0).attr('aria-label')).toBe('foo')
+    })
+
+    it('should have a primary next step button', () => {
```
jimmyfortinx: You are testing the `next` state, but what about the `previous` state, somebody could break it and we will never know ?

Same thing about `secondary` and `link` states.

packages/oui-button/src/index.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      expect(button.attr('disabled')).toBe('disabled')
+      expect(button.attr('type')).toBe('submit')
+    })
+
+    it('should call function of onClick attribute, when button is clicked', () => {
+      let element = angular.element('<oui-button text="foo" on-click="onClickTest()"></oui-button>')
+      let scope = $rootScope.$new()
+
+      scope.clickTest = false
+      scope.onClickTest = () => {
```
jimmyfortinx: You could use `jasmine.createSpy("onClick")` instead and use `expect(scope.onClickTest).toHaveBeenCalled()`.

packages/oui-button/src/button.html
```diff
@@ -0,0 +1,22 @@
...
+    'oui-button_secondary': $ctrl.variant === 'secondary',
+    'oui-button_link': $ctrl.variant === 'link',
+    'oui-button_icon-left': $ctrl.variantNav === 'previous',
+    'oui-button_icon-right': $ctrl.variantNav === 'next'
+  }"
+  ng-click="$ctrl.onClick()"
+  ng-disabled="$ctrl.disabled"
+>
+  <span aria-hidden="true" class="oui-icon oui-icon_circle oui-icon-chevron-left"
+    ng-if="$ctrl.variantNav === 'previous'"></span>
```
tremblaymath: One-time

packages/oui-button/src/button.html
```diff
@@ -0,0 +1,22 @@
...
+  <span ng-bind="::$ctrl.text"></span>
+  <span aria-hidden="true" class="oui-icon oui-icon_circle oui-icon-chevron-right"
+    ng-if="$ctrl.variantNav === 'next'"></span>
```
tremblaymath: one-time

packages/oui-button/src/index.spec.js
```diff
@@ -0,0 +1,109 @@
...
+    let $rootScope
+    let $compile
+
+    beforeEach(inject((_$rootScope_, _$compile_) => {
+      $rootScope = _$rootScope_
+      $compile = _$compile_
+    }))
+
+    it('should display a button with value of attribute text, and is type="button" by default', () => {
+      let element = angular.element('<oui-button text="foo"></oui-button>')
```
tremblaymath: This code is duplicated in all tests, we should extract a function e.g.: `getElement('<oui-button text="foo"></oui-button>')`
let element = angular.element('..')
let scope = $rootScope.$new()
$compile(element)(scope)
 scope.$digest()

packages/oui-button/src/button.html
AxelPeter: done

packages/oui-button/src/button.html
AxelPeter: done

packages/oui-button/src/index.spec.js
AxelPeter: Yes, simpler :)

packages/oui-button/src/index.spec.js
AxelPeter: I don't know. In this case, i'll have to test also `disabled="true"` and `disabled="false"` to be complete

Here, I prefer to test the attribute and its behaviour more than its value.
But why not if necessary :)

packages/oui-button/src/index.spec.js
AxelPeter: I'll use @neolitec 's helper and update it for my need

PR #18, opened 2017-11-17
feat(oui-numeric): add oui-numeric component.

frenautvh:
http://oui_numeric_staging.ui-kit.ovh/#!/oui-angular/numeric

### DIFF
--- a/packages/oui-numeric/README.md
+++ b/packages/oui-numeric/README.md
@@ -1,0 +1,27 @@
+# Numeric
+
+<component-status cx-design="complete" ux="prototype"></component-status>
+
+## Usage
+
+```html:preview
+<oui-numeric id="someId"
+             text="CPU"
+             unit="core"
+             model="$ctrl.cpuCores">
+</oui-numeric>
+```
+
+## API
+
+| Attribute     | Type     | Binding | Values              | Default | Description                         |
+| ----          | ----     | ----    | ----                | ----    | ----                                |
+| id            | string   | @?      |                     |         | id attribute of the input           |
+| name          | string   | @       |                     |         | name attribute of the input         |
+| text          | string   | @       |                     |         | label text                          |
+| unit          | string   | @?      |                     |         | unit displayed next to label        |
+| model         | string   | =       |                     |         | model bound to component            |
+| min           | interger | <?      |                     | 0       | model lower bound                   |
+| max           | interger | <?      |                     | 99999   | model upper bound                   |
+| disabled      | boolean  | <?      |                     | false   | disable the component               |
+| onChange      | function | &?      |                     |         | function to call when model changes |
--- a/packages/oui-numeric/README.md
+++ b/packages/oui-numeric/README.md
@@ -11,7 +11,7 @@
   model="..."
   min="..."
   max="..."
-  disabled|disabled="[true|false]"
+  disabled|disabled="true|false"
   on-change="...">
 </oui-numeric>
 ```
@@ -48,15 +48,6 @@
   name="foo"
   model="$ctrl.foo"
   disabled>
-</oui-numeric>
-```
-
-```html:preview
-<oui-numeric
-  id="bar"
-  name="bar"
-  model="$ctrl.bar"
-  disabled="true">
 </oui-numeric>
 ```
 
--- a/packages/oui-numeric/README.md
+++ b/packages/oui-numeric/README.md
@@ -18,7 +18,7 @@
 
 ## Examples
 
-#### Default
+### Default
 
 ```html:preview
 <oui-numeric
@@ -28,7 +28,7 @@
 </oui-numeric>
 ```
 
-#### Limits
+### Limits
 
 ```html:preview
 <oui-numeric
@@ -40,7 +40,7 @@
 </oui-numeric>
 ```
 
-#### Disabled
+### Disabled
 
 ```html:preview
 <oui-numeric
--- a/packages/oui-numeric/README.md
+++ b/packages/oui-numeric/README.md
@@ -14,14 +14,14 @@
 
 ## API
 
-| Attribute     | Type     | Binding | Values              | Default | Description                         |
-| ----          | ----     | ----    | ----                | ----    | ----                                |
-| id            | string   | @?      |                     |         | id attribute of the input           |
-| name          | string   | @       |                     |         | name attribute of the input         |
-| text          | string   | @       |                     |         | label text                          |
-| unit          | string   | @?      |                     |         | unit displayed next to label        |
-| model         | string   | =       |                     |         | model bound to component            |
-| min           | interger | <?      |                     | 0       | model lower bound                   |
-| max           | interger | <?      |                     | 99999   | model upper bound                   |
-| disabled      | boolean  | <?      |                     | false   | disable the component               |
-| onChange      | function | &?      |                     |         | function to call when model changes |
+| Attribute     | Type     | Binding | One-time binding | Values              | Default | Description                         |
+| ----          | ----     | ----    | ----             | ----                | ----    | ----                                |
+| id            | string   | @?      | true             |                     |         | id attribute of the input           |
+| name          | string   | @       | true             |                     |         | name attribute of the input         |
+| text          | string   | @       | true             |                     |         | label text                          |
+| unit          | string   | @?      | true             |                     |         | unit displayed next to label        |
+| model         | string   | =       |                  |                     |         | model bound to component            |
+| min           | integer  | <?      |                  |                     | 0       | model lower bound                   |
+| max           | integer  | <?      |                  |                     | 99999   | model upper bound                   |
+| disabled      | boolean  | <?      |                  |                     | false   | disable the component               |
+| onChange      | function | &?      |                  |                     |         | function to call when model changes |
--- a/packages/oui-numeric/README.md
+++ b/packages/oui-numeric/README.md
@@ -8,12 +8,10 @@
 <oui-numeric
   id="..."
   name="..."
-  text="..."
-  unit="..."
   model="..."
   min="..."
   max="..."
-  disabled:"[true|false]"
+  disabled|disabled="[true|false]"
   on-change="...">
 </oui-numeric>
 ```
@@ -26,8 +24,6 @@
 <oui-numeric
   id="partitionSize"
   name="partitionSize"
-  text="Disk size"
-  unit="Gb"
   model="$ctrl.size">
 </oui-numeric>
 ```
@@ -38,8 +34,6 @@
 <oui-numeric
   id="core"
   name="core"
-  text="CPU"
-  unit="core"
   min="2"
   max="8"
   model="$ctrl.core">
@@ -52,7 +46,6 @@
 <oui-numeric
   id="foo"
   name="foo"
-  text="Disbled"
   model="$ctrl.foo"
   disabled>
 </oui-numeric>
@@ -62,7 +55,6 @@
 <oui-numeric
   id="bar"
   name="bar"
-  text="Disbled"
   model="$ctrl.bar"
   disabled="true">
 </oui-numeric>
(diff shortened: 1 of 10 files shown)


packages/oui-numeric/README.md
```diff
@@ -0,0 +1,27 @@
...
+<oui-numeric id="someId"
+             text="CPU"
+             unit="core"
+             model="$ctrl.cpuCores">
+</oui-numeric>
+```
+
+## API
+
+| Attribute     | Type     | Binding | Values              | Default | Description                         |
```
AxelPeter: We should have a column for attributes with one-time bindings

See https://github.com/ovh-ux/ovh-ui-angular/tree/feature/oui-button/packages/oui-button (for example)

packages/oui-numeric/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Usage
+
+```html:preview
+<oui-numeric id="someId"
+             text="CPU"
+             unit="core"
+             model="$ctrl.cpuCores">
+</oui-numeric>
+```
```
AxelPeter: Maybe we should have more examples with `disabled`, `min`, `max`, etc. ?

packages/oui-numeric/src/numeric.html
```diff
@@ -0,0 +1,29 @@
+<label ng-attr-for="{{:: $ctrl.id }}"
```
AxelPeter: To be discussed on our next daily if we still need the prefix `data-`.
It matters only for the validation

packages/oui-numeric/src/numeric.html
```diff
@@ -0,0 +1,29 @@
...
+       class="oui-label">
+  <span ng-bind=":: $ctrl.text"></span>
+  <span ng-if="$ctrl.unit"
+        ng-bind=":: '&nbsp;(' + $ctrl.unit + ')'"></span>
```
AxelPeter: Is `&nbsp;` necessary ?
The 2 `span` are not glued, we'll have a space between

packages/oui-numeric/src/numeric.controller.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    if (!angular.isNumber(this.min)) {
+      if (angular.isDefined(this.min)) {
+        this.$log.warn(`Invalid attribute min, expected number got '${this.min}'`)
+      }
+      this.min = MIN_VALUE
+    }
+
+    if (!angular.isNumber(this.max)) {
```
AxelPeter: Maybe we could export that in a method like `setNumberAttr (attrName, defaultValue)` ?

We'll get:
``` 
this.setNumberAttr('min', MIN_VALUE)
this.setNumberAttr('max', MAX_VALUE)
```

packages/oui-numeric/src/numeric.controller.js
```diff
@@ -0,0 +1,117 @@
...
+      if (angular.isDefined(this.max)) {
+        this.$log.warn(`Invalid attribute max, expected number got '${this.max}'`)
+      }
+      this.max = MAX_VALUE
+    }
+
+    if (!angular.isNumber(this.model)) {
```
AxelPeter: See upper comment

packages/oui-numeric/src/index.spec.js
```diff
@@ -0,0 +1,155 @@
...
+      let scope = $rootScope.$new()
+      let tmp
+      scope.foo = 5
+      scope.onChange = (ev) => {
+        tmp = ev.value
+      }
+      $compile(elt)(scope)
+      scope.$digest()
+      elt.find('input').controller('ngModel').$setViewValue('10')
+      expect(tmp).toBe(10)
```
jimmyfortinx: You can use `expect(scope.onChange).toHaveBeenCalledWith(10)` if you use `jasmine.createSpy`

packages/oui-numeric/src/index.spec.js
```diff
@@ -0,0 +1,155 @@
...
+      let scope = $rootScope.$new()
+      let tmp
+      scope.foo = 5
+      scope.onChange = (ev) => {
+        tmp = ev.value
+      }
+      $compile(elt)(scope)
+      scope.$digest()
+      elt.find('input').controller('ngModel').$setViewValue('5')
+      expect(tmp).not.toBeDefined()
```
jimmyfortinx: `expect(scope.onChange).not.toHaveBeenCalled()` if you use `jasmine.createSpy`

packages/oui-numeric/src/numeric.component.js
```diff
@@ -0,0 +1,18 @@
...
+export default {
+  template,
+  controller,
+  bindings: {
+    id: '@',
+    name: '@?',
+    text: '@',
+    unit: '@?',
+    model: '=',
+    min: '<?',
```
jimmyfortinx: I would have used `@?` instead of `<` for `min` and `max`, what do you think ?

I think user will more use our component in this way: `<oui-numeric min="2" max="100"></oui-numeric>` than `<oui-numeric min="$ctrl.min" max="$ctrl.max"></oui-numeric>`

packages/oui-numeric/src/numeric.controller.js
```diff
@@ -0,0 +1,117 @@
+import _ from 'lodash'
```
jimmyfortinx: @neolitec You have integrated to tool where we can import lodash by part like `import { clamp } from 'lodash'` ?

packages/oui-numeric/src/numeric.controller.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  $postLink () {
+    this.$element.removeAttr('id')
+    this.$element.removeAttr('name')
+  }
+
+  $onInit () {
+    if (!this.id) {
+      this.$log.warn(`Missing required attribute id`)
```
jimmyfortinx: For checkbox if `id` is missing i'm generating one using `$scope.$id` like `oui-checkbox-0`. With that id will always be unique and `label for` will work without requiring the user to enter an id.

packages/oui-numeric/src/numeric.controller.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    if (!this.text) {
```
jimmyfortinx: Angular is not already throwing an error if `text` is missing because you are using `@` in bindings ?

packages/oui-numeric/src/numeric.controller.js
```diff
@@ -0,0 +1,117 @@
...
+      }
+      // if model is undefined, initialize it with min value
+      this.setModelValue(this.min)
+    }
+
+    if (this.min < MIN_VALUE) {
+      this.$log.warn(`Invalid attribute min, value should be greater than '${MIN_VALUE}'`)
+    }
+
+    if (this.max < MAX_VALUE) {
```
jimmyfortinx: wrong direction :) should be `this.max > MAX_VALUE`

You don't test your warn messages ;)

packages/oui-numeric/src/numeric.controller.js
```diff
@@ -0,0 +1,117 @@
...
+      this.$log.warn(`Invalid attribute max, value should be lower than '${MAX_VALUE}'`)
+    }
+
+    if (angular.isDefined(this.$attrs.disabled) && angular.isUndefined(this.disabled)) {
+      this.disabled = true
+    }
+
+    this.min = _.clamp(this.min, MIN_VALUE, MAX_VALUE)
+    this.max = _.clamp(this.max, MIN_VALUE, MAX_VALUE)
```
jimmyfortinx: Should it be `_.clamp(this.max, this.min, MAX_VALUE)` ?

packages/oui-numeric/src/numeric.html
```diff
@@ -0,0 +1,31 @@
...
+  <span ng-bind=":: $ctrl.text"></span>
+  <span ng-if="$ctrl.unit"
+        ng-bind=":: '(' + $ctrl.unit + ')'"></span>
+</label>
+<div class="oui-input-group oui-input-group_numeric">
+  <button class="oui-button oui-button_icon-only oui-button_small-width"
+          type="button"
+          ng-disabled="$ctrl.disabled || $ctrl.model <= $ctrl.min"
+          ng-click="$ctrl.decrement()"
+          tabindex="-1"
```
jimmyfortinx: We really want to never tab on those buttons ?

packages/oui-numeric/src/numeric.controller.js
neolitec: `babel-plugin-lodash` in combination with `lodash-webpack-plugin` can help cleaning the dist of unused lodash parts. It's already in the project.

packages/oui-numeric/src/numeric.controller.js
frenautvh: No it doesn't, value will be undefined

packages/oui-numeric/src/numeric.html
frenautvh: I think so, if you use keyboard and tab navigate to the component you will directly input the value. Then the next natural action will be to tab out to the next field. Buttons here are meant to be used with the mouse imho

packages/oui-numeric/src/numeric.controller.js
frenautvh: I really don't this idea. First, i'm not setting an attribute here, i'm checking if this.max is a number and in case it's not i log the issue and defaults its value to MAX_VALUE.
What would you want, some kind of "checkIfNumberOrDefaultsTo()" function ? :)

Such simple logic doesn't need to be refactored in a function. IMHO the function will be hiding the control flow leading to less code readability. Feels free to comment if you disagree ;)

packages/oui-numeric/src/index.spec.js
```diff
@@ -0,0 +1,155 @@
...
+      $compile(elt)(scope)
+      scope.$digest()
+      expect(elt.find('input').val()).toBe('25')
+      scope.foo = 5
+      scope.$digest()
+      expect(elt.find('input').val()).toBe('5')
+    })
+
+    it('should not updates model when input is given invalid value', () => {
+      let elt = angular.element('<oui-numeric model="foo" min="0" max="5"></oui-numeric>')
```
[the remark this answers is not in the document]
frenautvh: yes i could refactor these 2-3 lines into a function that returns { element: ... scope: ... } but i don't think we gain much more code readability by doing so, maybe we'll just save a couple of lines. But if you insist (and maybe send me a cookie) i'll do it

packages/oui-numeric/src/numeric.controller.js
AxelPeter: It was not just for the checks, but also for the warning message. Same in the 3 cases.
It's not like we do a method just for one `if` :)

packages/oui-numeric/src/numeric.html
AxelPeter: Yes, not necessary. 
On type number, you can do with the keyboard what the buttons do (up and down).
ex. https://www.w3schools.com/html/tryit.asp?filename=tryhtml_input_number

jimmyfortinx: Good point about the up and down key, didn't think about that :)

packages/oui-numeric/README.md
```diff
@@ -0,0 +1,83 @@
...
+```html
+<oui-numeric
+  id="..."
+  name="..."
+  text="..."
+  unit="..."
+  model="..."
+  min="..."
+  max="..."
+  disabled:"[true|false]"
```
antleblanc: `disabled|disabled="true|false"` (same has #13)

packages/oui-numeric/README.md
```diff
@@ -0,0 +1,83 @@
...
+</oui-numeric>
+```
+
+#### Disabled
+
+```html:preview
+<oui-numeric
+  id="foo"
+  name="foo"
+  text="Disbled"
```
antleblanc: `Disbled` => `Disabled`?

packages/oui-numeric/README.md
```diff
@@ -0,0 +1,83 @@
...
+  model="$ctrl.foo"
+  disabled>
+</oui-numeric>
+```
+
+```html:preview
+<oui-numeric
+  id="bar"
+  name="bar"
+  text="Disbled"
```
antleblanc: `Disbled` => `Disabled`?

PR #21, opened 2017-11-22
feat(oui-tooltip): add oui-tooltip component

AxelPeter:
UU-519 (linked to https://github.com/ovh-ux/ovh-ui-kit/pull/104)

Preview http://temp_oui_tooltip.ui-kit.ovh

### DIFF
--- a/packages/oui-tooltip/src/tooltip.controller.js
+++ b/packages/oui-tooltip/src/tooltip.controller.js
@@ -42,17 +42,7 @@ export default class {
 
     // Let Popper.js position the tooltip
     this.popper = new Popper(trigger, tooltip, {
-      placement: this.placement,
-      onCreate: popper => this.updatePlacement(popper),
-      onUpdate: popper => this.updatePlacement(popper)
+      placement: this.placement
     })
-  }
-
-  updatePlacement (popper) {
-    // Update placement if flipped
-    if (this.flipped !== popper.flipped) {
-      this.flipped = popper.flipped
-      this.$scope.$apply()
-    }
   }
 }
--- a/packages/oui-tooltip/src/tooltip.controller.js
+++ b/packages/oui-tooltip/src/tooltip.controller.js
@@ -1,0 +1,58 @@
+import template from './tooltip.html'
+import Popper from 'popper.js'
+
+export default class {
+  constructor ($compile, $element, $scope) {
+    'ngInject'
+
+    this.$compile = $compile
+    this.$element = $element
+    this.$scope = $scope
+  }
+
+  $onInit () {
+    // Add default value for attribute 'placement'
+    if (angular.isUndefined(this.placement)) {
+      this.placement = 'top'
+    }
+  }
+
+  $postLink () {
+    // Add an attribute 'aria-label' if undefined
+    if (!this.$element.attr('aria-label')) {
+      this.$element.attr('aria-label', this.text)
+    }
+
+    // Create a new scope to compile the tooltip next to the trigger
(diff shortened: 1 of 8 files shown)


packages/oui-tooltip/src/tooltip.html
```diff
@@ -0,0 +1,13 @@
+<div class="oui-tooltip oui-message" role="tooltip">
+  <span aria-hidden="true"
+    ng-class="{
```
neolitec: Is this part can be simplified in a method?
It will be simplified when if it remains only top|bottom|left|right , but still...

Another point: one-time binding maybe?

packages/oui-tooltip/src/tooltip.html
```diff
@@ -0,0 +1,13 @@
...
+      'oui-tip-top':                  (!$ctrl.flipped && $ctrl.placement === 'top') || ($ctrl.flipped && $ctrl.placement === 'bottom'),
```
neolitec: Is this part can be simplified in a method?
It will be simplified when if it remains only top|bottom|left|right , but still...
If I read the guidelines:

> Default opening direction is "top center"
The content opens inside the viewport in the direction where there is the most available space.

So there is no place for a "placement" attribute. (I think you should aven don't use end/start suffixes in the case of the tooltips).

Another point: one-time binding maybe?

packages/oui-tooltip/src/tooltip.controller.js
```diff
@@ -0,0 +1,58 @@
...
+
+    // Let Popper.js position the tooltip
+    this.popper = new Popper(trigger, tooltip, {
+      placement: this.placement,
+      onCreate: popper => this.updatePlacement(popper),
+      onUpdate: popper => this.updatePlacement(popper)
+    })
+  }
+
+  updatePlacement (popper) {
```
neolitec: arrow position can be defined with the attribute x-placement (see dropdown)

packages/oui-tooltip/src/tooltip.controller.js
```diff
@@ -0,0 +1,58 @@
...
+
+    // Create a new scope to compile the tooltip next to the trigger
+    const tooltipScope = angular.extend(this.$scope.$new(true), { $ctrl: this })
+    const tooltipTemplate = this.$compile(template)(tooltipScope)
+
+    this.$element
+      // Add classname for 'focus' and 'hover' CSS events
+      .addClass('oui-tooltip__trigger')
+      // one time bind to create the popper helper
+      .one('focus mouseenter', () => this.createPopper())
```
neolitec: I you leave the button with tab key, the previous tooltip is still visible.

packages/oui-tooltip/src/tooltip.controller.js
AxelPeter: Yes, it's better now :)

packages/oui-tooltip/src/tooltip.controller.js
AxelPeter: Ah ? 
I don't have the problem with the Tab navigation :/

packages/oui-tooltip/src/tooltip.html
AxelPeter: I'll look to try a simpler way to do that. But now with the management of the arrow by Popper, it'll be.

If one-time, i can't change it when flipped.

packages/oui-tooltip/src/tooltip.controller.js
jimmyfortinx: Tested and it is related to tab and mouse hover events triggered but saw with CX and this behavior is not something that we need to fix.

PR #22, opened 2017-11-23
feat(oui-back-button): add oui-back-button component.

frenautvh:
http://angular_feat_back_button.ui-kit.ovh/#!/oui-angular/back-button


packages/oui-back-button/README.md
```diff
@@ -0,0 +1,42 @@
+# Back button
+
+<component-status cx-design="complete" ux="prototype"></component-status>
+
+## Usage
+
+```html
+<oui-back-button>
```
jimmyfortinx: Need to be updated to list all attributes.

packages/oui-back-button/README.md
```diff
@@ -0,0 +1,42 @@
...
+</oui-back-button>
+```
+
+## Examples
+
+### Default
+
+```html:preview
+<oui-back-button>
```
jimmyfortinx: The back button (on CX specs) also includes the text at the right, do we want it ?

packages/oui-back-button/src/index.spec.js
```diff
@@ -0,0 +1,65 @@
...
+  let $compile
+  let $window
+
+  beforeEach(angular.mock.module('oui.back-button'))
+
+  beforeEach(inject((_$rootScope_, _$compile_, _$window_) => {
+    $rootScope = _$rootScope_
+    $compile = _$compile_
+    $window = _$window_
+    $window.history = {
```
jimmyfortinx: You should use `spyOn($window.history, 'back')` instead

packages/oui-back-button/README.md
neolitec: Too general name for button that must be placed at the left of an h2 top positioned title I think.

frenautvh: if we had the h2 title yes it makes sense to rename it

packages/oui-back-button/README.md
frenautvh: not sure about this one, do we really want to restrict the component to be used with a h2? (similar to : do we add labels to input components)
if that's the case, it won't be a oui-back-button anymore but maybe something like oui-page-title (please give me great ideas ;) )

jimmyfortinx: Saw together, we will keep it simple.

PR #27, opened 2017-11-28
feat(oui-checkbox): description attribute

jimmyfortinx:
Signed-off-by: Jimmy Fortin <jimmy.fortin@corp.ovh.com>

http://angular_feat_checkbox_description.ui-kit.ovh/#!/oui-angular/checkbox

### DIFF
--- a/packages/oui-checkbox/src/checkbox.html
+++ b/packages/oui-checkbox/src/checkbox.html
@@ -8,6 +8,7 @@
          ng-disabled="$ctrl.disabled">
   <label for="{{:: $ctrl.id }}" class="oui-checkbox__label-container">
     <span class="oui-checkbox__label" ng-bind="$ctrl.text"></span>
+    <span class="oui-checkbox__description" ng-bind="$ctrl.description" ng-if="$ctrl.description"></span>
     <span class="oui-checkbox__icon">
       <i class="oui-icon oui-icon-checkbox-unchecked" aria-hidden="true"></i>
       <i class="oui-icon oui-icon-checkbox-checked" aria-hidden="true"></i>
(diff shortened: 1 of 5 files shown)


packages/oui-checkbox/src/checkbox.html
```diff
@@ -8,6 +8,7 @@
          ng-disabled="$ctrl.disabled">
   <label for="{{:: $ctrl.id }}" class="oui-checkbox__label-container">
     <span class="oui-checkbox__label" ng-bind="$ctrl.text"></span>
+    <span class="oui-checkbox__description" ng-bind="$ctrl.description" ng-if="$ctrl.description"></span>
```
jleveugle: The label must be explicit and concise, 'cause the label is read when the linked input is focus.

So you will hear : "Label Description BLABLABLABLABLABLABLA - Text field - Required"

You should consider aria-describedby and keep a simple label :)

AxelPeter: Maybe with a structure like:
```
<div>
  <input>
  <label></label>
  <span>></label>
</div>
```

AxelPeter: Forget my upper comment,
https://developer.mozilla.org/fr/docs/Accessibilit%C3%A9/ARIA/Techniques_ARIA/Utiliser_l_attribut_aria-describedby

Like said @jleveugle `aria-labelledby` and `aria-describedby` seems to resolve this issue.

jimmyfortinx: Fixed, I used `aria-labelledby` and `aria-describedby` since we need the description to be clickable.

PR #28, opened 2017-11-29
fix(oui-numeric): fix accessibility attributes.

frenautvh:
replace wrong accessiblity attributes

http://angular_fix_oui_numeric_accessibility.ui-kit.ovh/#!/oui-angular/numeric

### DIFF
--- a/packages/oui-numeric/src/numeric.html
+++ b/packages/oui-numeric/src/numeric.html
@@ -4,13 +4,17 @@
           ng-disabled="$ctrl.disabled || $ctrl.model <= $ctrl.min"
           ng-click="$ctrl.decrement()"
           tabindex="-1"
-          aria-labelledby="-">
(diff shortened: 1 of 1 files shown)


packages/oui-numeric/src/numeric.html
```diff
@@ -4,13 +4,17 @@
           ng-disabled="$ctrl.disabled || $ctrl.model <= $ctrl.min"
           ng-click="$ctrl.decrement()"
           tabindex="-1"
-          aria-labelledby="-">
+          aria-hidden="true">
     <i class="oui-icon oui-icon-remove" aria-hidden="true"></i>
```
AxelPeter: (optionnal)

`aria-hidden` still necessary, if the parent has it ?

frenautvh: you're right, i forgot there was an aria-hidden on <i>, will remove it

PR #29, opened 2017-11-29
UU-551 - Escape button does not close the action menu

neolitec:

http://angular_uu_551_fix_esc_dropdown.ui-kit.ovh/#!/oui-angular/action-menu

### DIFF
--- a/packages/oui-dropdown/src/dropdown-trigger.directive.js
+++ b/packages/oui-dropdown/src/dropdown-trigger.directive.js
@@ -34,11 +34,18 @@ export default ($compile, $document) => {
 
             triggerElement.on("click", () => ctrl.onTriggerClick());
             triggerElement.on("blur", evt => ctrl.triggerBlurHandler(evt));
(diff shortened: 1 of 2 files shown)


packages/oui-dropdown/src/dropdown-trigger.directive.js
```diff
@@ -33,8 +33,8 @@ export default $compile => {
             triggerElement.attr("id", ctrl.id);
 
             triggerElement.on("click", () => ctrl.onTriggerClick());
-            triggerElement.on("keydown", evt => ctrl.triggerKeyHandler(evt));
             triggerElement.on("blur", evt => ctrl.triggerBlurHandler(evt));
+            $document.on("keydown", evt => ctrl.triggerKeyHandler(evt));
```
jimmyfortinx: (optional) Could it be registered only when the dropdown is expanded ? Because here if we have multiple instance of the trigger each time keydown will pass through those handler even if no dropdown is shown, unless I missed something.

PR #32, opened 2017-12-04
fix(oui-checkbox): element structure

jimmyfortinx:
Signed-off-by: Jimmy Fortin <jimmy.fortin@corp.ovh.com>

http://input_height_staging.ui-kit.ovh/#!/oui-angular/checkbox

### DIFF
--- a/packages/oui-checkbox/src/index.spec.js
+++ b/packages/oui-checkbox/src/index.spec.js
@@ -131,7 +131,7 @@ describe("ouiCheckbox", () => {
                 };
 
                 const element = TestUtils.compileTemplate("<oui-checkbox model=\"$ctrl.currentModel\"></oui-checkbox>", context);
-                const $ctrl = TestUtils.constructor.getElementController(element);
+                const $ctrl = TestUtils.getElementController(element);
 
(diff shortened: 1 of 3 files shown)


packages/oui-checkbox/src/index.spec.js
```diff
@@ -131,7 +131,7 @@ describe("ouiCheckbox", () => {
                 };
 
                 const element = TestUtils.compileTemplate("<oui-checkbox model=\"$ctrl.currentModel\"></oui-checkbox>", context);
-                const $ctrl = TestUtils.getElementController(element);
+                const $ctrl = TestUtils.constructor.getElementController(element);
```
neolitec: Very ugly. VERY I mean.  An exception should be made for this case.

jimmyfortinx: Do you prefer that I export the `TestUtils` class and I import it under `TestUtilsClass` so I could do `TestUtilsClass.getElementController(element)` instead ?

jimmyfortinx: I applied the eslint-disable comment, but if it happens too much we need to find a better solution.

PR #33, opened 2017-12-04
fix(actionMenu): fix action menu positioning

neolitec:
Fixes UU-582

http://angular_uu_582_action_menu_in_datagrid.ui-kit.ovh/#!/oui-angular/action-menu

### DIFF
--- a/packages/oui-action-menu/src/action-menu.html
+++ b/packages/oui-action-menu/src/action-menu.html
@@ -1,4 +1,8 @@
-<oui-dropdown align="{{::$ctrl.align}}"
+<!--
+  Use data-align to avoid conflict with native align attribute.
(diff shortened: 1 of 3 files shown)


packages/oui-action-menu/src/action-menu.html
```diff
@@ -1,4 +1,4 @@
-<oui-dropdown align="{{::$ctrl.align}}"
+<oui-dropdown data-align="{{::$ctrl.align}}"
```
jimmyfortinx: Why add `data-` only on `align` attribute and not also on `arrow` on this case ? Is it the root cause of the problem you were fixing ?

neolitec: "Fixed" with a comment.

PR #39, opened 2017-12-08
feat(pagination): add new component

neolitec:

### DIFF
--- a/packages/oui-pagination/README.md
+++ b/packages/oui-pagination/README.md
@@ -1,0 +1,63 @@
+# Pagination
+
+<component-status cx-design="complete" ux="prototype"></component-status>
+
+
+
+## Usage
+
+### A few pages
+
+```html:preview
+<oui-pagination
+    current-offset="$ctrl.pagination1.offset"
+    total-items="$ctrl.pagination1.totalItems"
+    on-change="$ctrl.onChange('pagination1', $event)">
+    </oui-pagination>
+```
+
+### Lots of pages
+
+```html:preview
+<oui-pagination
+    current-offset="$ctrl.pagination2.offset"
+    page-size="$ctrl.pagination2.pageSize"
+    total-items="$ctrl.pagination2.totalItems"
+    on-change="$ctrl.onChange('pagination2', $event)">
+    </oui-pagination>
+```
+
+## API
+
+| Attribute         | Type     | Binding | One-time Binding | Values                 | Default             | Description                                      |
+| ----              | ----     | ----    | ----             | ----                   | ----                | ----                                             |
+| `current-offset`  | Number   | <       | no               |                        |                     | offset of the current page first item            |
+| `page-size`       | Number   | <?      | no               |                        | defined by provider | number of items per page                         |
+| `page-size-max`   | Number   | <?      | no               |                        |                     | max page size of the page sizes list             |
+| `total-items`     | Number   | <       | no               |                        |                     | total number of items                            |
+| `on-change`       | function | &?      |                  |                        |                     | change callback, called on pagination changes    |
+
+`on-change` callback takes a parameter `$event` containing:
+
+  - `offset`: the new offset
+  - `pageSize`: the new page size
+
+## Configuration
+
+The pagination can be globally configured with a provider.
+
+```js
+angular.module("myModule", [
+    "oui.pagination"
+]).config(ouiPaginationConfigurationProvider => {
+    ouiPaginationConfigurationProvider.setPageSize(25); // default page size (when page-size attribute is not set)
+    ouiPaginationConfigurationProvider.setPageSizeList([25, 50, 100, 300]); // List of page sizes
+    ouiPaginationConfigurationProvider.setTranslations({ // Translations (double curly braces for placeholders)
+        resultsPerPage: "Results per page",
+        ofNResults: "of {{totalItems}} results",
+        currentPageOfPageCount: "Page {{currentPage}} of {{pageCount}}"
+    });
+});
+```
+
+
--- a/packages/oui-pagination/README.md
+++ b/packages/oui-pagination/README.md
@@ -27,6 +27,16 @@
     </oui-pagination>
 ```
 
+### One page
+
+```html:preview
+<oui-pagination
+    current-offset="1"
+    page-size="100"
+    total-items="12">
+    </oui-pagination>
+```
+
 ## API
 
 | Attribute         | Type     | Binding | One-time Binding | Values                 | Default             | Description                                      |
@@ -55,7 +65,9 @@ angular.module("myModule", [
     ouiPaginationConfigurationProvider.setTranslations({ // Translations (double curly braces for placeholders)
         resultsPerPage: "Results per page",
         ofNResults: "of {{totalItems}} results",
-        currentPageOfPageCount: "Page {{currentPage}} of {{pageCount}}"
+        currentPageOfPageCount: "Page {{currentPage}} of {{pageCount}}",
+        previousPage: "Previous page",
+        nextPage: "Next page"
     });
 });
 ```
--- a/packages/oui-pagination/src/index.spec.js
+++ b/packages/oui-pagination/src/index.spec.js
@@ -292,6 +292,28 @@ describe("ouiPagination", () => {
                 expect(pageButtons.length).toEqual(20);
             });
 
+            it("should reset offset to 1 when page size is changed", () => {
+                const clickSpy = jasmine.createSpy("clickSpy");
+                const element = TestUtils.compileTemplate(`
+                    <oui-pagination
+                        current-offset="999"
+                        page-size="$ctrl.pageSize"
+                        total-items="1000"
+                        on-change="$ctrl.clickHandler($event)">
+                    </oui-pagination>
+                `, {
+                        clickHandler: clickSpy
+                    });
+
+                const pageSizesButtons = getProgress(element).querySelectorAll(".oui-pagination-menu__items-list .oui-pagination-menu__item");
+                angular.element(pageSizesButtons[1]).triggerHandler("click");
+
+                expect(clickSpy).toHaveBeenCalledWith(jasmine.objectContaining({ offset: 1 }));
+                expect(clickSpy.calls.count()).toEqual(1);
+                const paginationController = element.controller("ouiPagination");
+                expect(paginationController.currentOffset).toEqual(1);
+            });
+
             it("should have max page size limiting the page size list (without creating a new value in this list)", () => {
                 const element = TestUtils.compileTemplate(`
                     <oui-pagination
--- a/packages/oui-pagination/src/index.spec.js
+++ b/packages/oui-pagination/src/index.spec.js
@@ -1,0 +1,336 @@
+describe("ouiPagination", () => {
+    let TestUtils;
+
+    const getPagination = elt => elt[0].querySelector(".oui-pagination");
+    const getProgress = elt => elt[0].querySelector(".oui-pagination__progress");
+    const getSelector = elt => elt[0].querySelector(".oui-pagination__selector");
+    const getPreviousButton = elt => elt[0].querySelectorAll(".oui-pagination__selector > button")[0];
+    const getNextButton = elt => elt[0].querySelectorAll(".oui-pagination__selector > button")[1];
+    const getPageSizeButtons = elt => getProgress(elt).querySelectorAll(".oui-pagination-menu__items-list .oui-pagination-menu__item");
+
+    beforeEach(angular.mock.module("oui.pagination"));
+    beforeEach(angular.mock.module("oui.test-utils"));
+    beforeEach(angular.mock.module("test.paginationConfiguration"));
+
+    beforeEach(inject((_TestUtils_) => {
+        TestUtils = _TestUtils_;
+    }));
+
+    describe("Provider", () => {
+        let paginationConfiguration;
+
+        angular.module("test.paginationConfiguration", [
+            "oui.pagination"
+        ]).config(ouiPaginationConfigurationProvider => {
+            ouiPaginationConfigurationProvider.setPageSize(50);
+            ouiPaginationConfigurationProvider.setPageSizeList([25, 50, 100, 200]);
+            ouiPaginationConfigurationProvider.setTranslations({
+                resultsPerPage: "Results per page",
+                ofNResults: "of {{totalItems}} results",
+                currentPageOfPageCount: "Page {{currentPage}} of {{pageCount}}",
+                foo: "bar"
+            });
+        });
+
+        beforeEach(inject(_ouiPaginationConfiguration_ => {
+            paginationConfiguration = _ouiPaginationConfiguration_;
+        }));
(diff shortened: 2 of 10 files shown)


packages/oui-pagination/src/pagination.html
```diff
@@ -0,0 +1,77 @@
...
+                <span ng-bind="$ctrl.translations.currentPageOfPageCount"></span>
+                <i class="oui-icon oui-icon-chevron-down" aria-hidden="true"></i>
+            </button>
+            <oui-dropdown-content>
+                <div role="menu" class="oui-pagination-menu">
+                    <div class="oui-pagination-menu__items-list">
+                        <button type="button"
+                            ng-repeat="page in $ctrl.pageRange track by page"
+                            class="oui-pagination-menu__item"
+                            ng-disabled="page === $ctrl.currentPage"
```
jimmyfortinx: (optional) I think we will need to add an `aria-label` on those buttons too otherwise the client will only understand the number instead of **page 2 of 5**.

packages/oui-pagination/src/pagination.html
```diff
@@ -0,0 +1,77 @@
...
+                            ng-bind="page"
+                            ng-click="$ctrl.onPageChange(page)"></button>
+                    </div>
+                </div>
+            </oui-dropdown-content>
+        </oui-dropdown>
+        <button type="button" class="oui-button oui-button_secondary oui-button_icon-only oui-button_small-width"
+            type="button"
+            aria-label="Next page"
```
jimmyfortinx: Hard coded `aria-label`, should be a config or it should not be set.

packages/oui-pagination/src/pagination.controller.js
```diff
@@ -0,0 +1,76 @@
...
+        this.translations.ofNResults = this.translations.ofNResults
+            .replace("{{totalItems}}", this.totalItems);
+        this.translations.currentPageOfPageCount = this.translations.currentPageOfPageCount
+            .replace("{{currentPage}}", this.currentPage)
+            .replace("{{pageCount}}", this.pageCount);
+    }
+
+    _onChange () {
+        this.onChange({
+            $event: {
```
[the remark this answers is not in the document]
neolitec: We was talking about modelValue specifically.
Here, this eventually multiplies available arguments in callback. I don't like that, so I propose using $event when we have multiple values that are not a single modelValue.

packages/oui-pagination/src/pagination.controller.js
```diff
@@ -0,0 +1,76 @@
...
+    getPageCount () {
+        return Math.ceil(this.totalItems / this.pageSize);
+    }
+
+    getCurrentPage () {
+        return Math.floor((this.currentOffset - 1) / this.pageSize) + 1;
+    }
+
+    getPageRange () {
+        return Array(...{ length: this.getPageCount() })
```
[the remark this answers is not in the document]
neolitec: No, not the same thing.

packages/oui-pagination/src/index.spec.js
```diff
@@ -0,0 +1,325 @@
...
+                    </oui-pagination>
+                    `, {
+                        clickHandler: clickSpy
+                    });
+
+                angular.element(getNextButton(element)).triggerHandler("click");
+                expect(clickSpy).toHaveBeenCalled();
+                expect(clickSpy.calls.count()).toEqual(1);
+
+                const $event = clickSpy.calls.mostRecent().args[0];
```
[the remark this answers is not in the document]
neolitec: No because toHaveBeenCalledWith tests a strict equality. This is not the case, there is `pageSize` next to `offset`.

packages/oui-pagination/src/index.spec.js
```diff
@@ -0,0 +1,325 @@
...
+                let $event;
+
+                expect(buttons[0].hasAttribute("disabled")).toBeTruthy();
+
+                angular.element(buttons[1]).triggerHandler("click");
+                expect(clickSpy).toHaveBeenCalled();
+                expect(clickSpy.calls.count()).toEqual(1);
+                $event = clickSpy.calls.mostRecent().args[0];
+                expect($event).toBeDefined();
+                expect($event.offset).toEqual(26);
```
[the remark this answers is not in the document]
neolitec: No, not the same buttons!

packages/oui-pagination/src/index.spec.js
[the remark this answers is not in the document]
jimmyfortinx: So, what you want is that:

```javascript
expect(clickSpy).toHaveBeenCalledWith(jasmine.objectContaining({ offset: 26 }));
```

packages/oui-pagination/src/pagination.controller.js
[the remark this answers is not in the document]
jimmyfortinx: Ok, I confirm it is not the same thing, it produce an array with the right number of item, but they can't be used. We need to execute the `fill` method after, otherwise we can't iterator over those elements. 

`Array(5).map((value, index) => index + 1)` produce `[ , , , , ]` instead of `[ 1, 2, 3, 4, 5]`

So, `Array(this.getPageCount()).fill().map(...)` is really equivalent of what you have done. I still don't really know why `...{ length: [...] }` seems to work, but it is OK for me (didn't found any documentation about it).

packages/oui-pagination/src/index.spec.js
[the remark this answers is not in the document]
jimmyfortinx: Good point

packages/oui-pagination/src/index.spec.js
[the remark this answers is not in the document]
neolitec: I forgot this technique, thx!

packages/oui-pagination/src/pagination.controller.js
```diff
@@ -0,0 +1,76 @@
...
+    constructor (ouiPaginationConfiguration) {
+        "ngInject";
+
+        this.config = ouiPaginationConfiguration;
+        this.pageSizeList = this.config.pageSizeList;
+    }
+
+    $onChanges (changes) {
+        // If pageSizeMax changes, recalculate pageSizeList...
+        if (changes.pageSizeMax && this.pageSizeMax) {
```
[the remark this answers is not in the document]
neolitec: Fixed

packages/oui-pagination/src/pagination.component.js
```diff
@@ -0,0 +1,14 @@
+import controller from "./pagination.controller";
+import template from "./pagination.html";
+
+export default {
+    template,
+    bindings: {
+        currentOffset: "<",
+        pageSize: "<",
+        pageSizeMax: "<",
```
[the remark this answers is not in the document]
neolitec: done

packages/oui-pagination/src/pagination.html
neolitec: Fixed

packages/oui-pagination/src/pagination.html
neolitec: Fixed

packages/oui-pagination/src/pagination.controller.js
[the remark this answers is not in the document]
jimmyfortinx: Ok for me. @AxelPeter what do you think about it ? If we all agree I'll let @neolitec update the `decisions.md` file with that.

packages/oui-pagination/README.md
```diff
@@ -0,0 +1,63 @@
...
+```js
+angular.module("myModule", [
+    "oui.pagination"
+]).config(ouiPaginationConfigurationProvider => {
+    ouiPaginationConfigurationProvider.setPageSize(25); // default page size (when page-size attribute is not set)
+    ouiPaginationConfigurationProvider.setPageSizeList([25, 50, 100, 300]); // List of page sizes
+    ouiPaginationConfigurationProvider.setTranslations({ // Translations (double curly braces for placeholders)
+        resultsPerPage: "Results per page",
+        ofNResults: "of {{totalItems}} results",
+        currentPageOfPageCount: "Page {{currentPage}} of {{pageCount}}"
```
tremblaymath: You could add these translations
previousPage: "Previous page",
nextPage: "Next page"

packages/oui-pagination/src/index.spec.js
```diff
@@ -0,0 +1,336 @@
...
+                angular.element(getPreviousButton(element)).triggerHandler("click");
+                expect(clickSpy).toHaveBeenCalledWith({
+                    offset: 51,
+                    pageSize: 25
+                });
+                expect(clickSpy.calls.count()).toEqual(1);
+            });
+        });
+
+        describe("Pages buttons", () => {
```
tremblaymath: It's not easy to read and it doesn't document the page buttons behavior.
e.g.
"should trigger on-change handler when a new page is selected" It should be separated in 3 distinct blocks
- Arrange template and mocks
- Act : Click a page button
- Assert :  Check on-change params

"should disable button of the current page" It should be separated in 3 distinct blocks
- Arrange template and mocks
- Act : Click a page button
- Assert :  Check attribute

...

When I read buttons[1], I don't know what it is. page1Button, page2Button could be easier to understand

packages/oui-pagination/src/index.spec.js
```diff
@@ -0,0 +1,336 @@
...
+                expect(clickSpy).toHaveBeenCalledWith({
+                    offset: 76,
+                    pageSize: 25
+                });
+                expect(clickSpy.calls.count()).toEqual(1);
+            });
+        });
+
+        describe("Pages dropdown", () => {
+            it("should change page", () => {
```
tremblaymath: should change page.. and what's the behavior (disable element, trigger, etc)

packages/oui-pagination/src/index.spec.js
```diff
@@ -0,0 +1,336 @@
...
+                        total-items="1000">
+                    </oui-pagination>
+                `);
+
+                const buttons = getProgress(element).querySelectorAll(".oui-pagination-menu__items-list .oui-pagination-menu__item");
+                expect(buttons.length).toEqual(4);
+                expect(buttons[0].hasAttribute("disabled")).toBeTruthy();
+            });
+
+            it("should change page size", () => {
```
tremblaymath: same things, 3 different tests I thing

packages/oui-pagination/src/index.spec.js
```diff
@@ -0,0 +1,336 @@
...
+                    <oui-pagination
+                        current-offset="1"
+                        page-size="25"
+                        page-size-max="50"
+                        total-items="1000">
+                    </oui-pagination>
+                `);
+
+                const pageSizesButtons = getProgress(element).querySelectorAll(".oui-pagination-menu__items-list .oui-pagination-menu__item");
+                expect(pageSizesButtons.length).toEqual(2);
```
tremblaymath: It's not easy to know why it's 2(magic number). expect(pageSizesButtons.length).toEqual(PageSizeMax/pageSize) could be easier to understand

packages/oui-pagination/README.md
neolitec: Yes, not documented yet.

packages/oui-pagination/src/index.spec.js
```diff
@@ -0,0 +1,336 @@
...
+                const contextController = element.scope().$ctrl;
+
+                expect(pageSizesButtons.length).toEqual(3);
+                expect(pageSizesButtons[2].innerHTML).toEqual("80");
+
+                // Reset to all page sizes
+                delete contextController.pageSizeMax;
+                element.scope().$digest();
+                pageSizesButtons = getPageSizeButtons(element);
+                expect(pageSizesButtons.length).toEqual(4);
```
tremblaymath: Not easy to understand, pageSizeMax is forced to 80, after it's deleted and page size 200 is now available.

packages/oui-pagination/src/pagination.controller.js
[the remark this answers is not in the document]
tremblaymath: ok for me too

packages/oui-pagination/src/pagination.html
```diff
@@ -25,7 +25,8 @@
         <span ng-bind="$ctrl.translations.ofNResults"></span>
     </div>
 
-    <div class="oui-pagination__selector">
+    <div class="oui-pagination__selector"
+        ng-if="$ctrl.totalItems > $ctrl.pageSize">
```
tremblaymath: A test should be added for this case

packages/oui-pagination/src/index.spec.js
```diff
@@ -0,0 +1,336 @@
...
+    describe("Provider", () => {
+        let paginationConfiguration;
+
+        angular.module("test.paginationConfiguration", [
+            "oui.pagination"
+        ]).config(ouiPaginationConfigurationProvider => {
+            ouiPaginationConfigurationProvider.setPageSize(50);
+            ouiPaginationConfigurationProvider.setPageSizeList([25, 50, 100, 200]);
+            ouiPaginationConfigurationProvider.setTranslations({
+                resultsPerPage: "Results per page",
```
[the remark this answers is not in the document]
neolitec: If I don't set theses values, all tests fail. I'd rather set realistic values than something else.

packages/oui-pagination/src/index.spec.js
```diff
@@ -0,0 +1,336 @@
...
+                currentPageOfPageCount: "Page {{currentPage}} of {{pageCount}}",
+                foo: "bar"
+            });
+        });
+
+        beforeEach(inject(_ouiPaginationConfiguration_ => {
+            paginationConfiguration = _ouiPaginationConfiguration_;
+        }));
+
+        it("should have custom values", () => {
```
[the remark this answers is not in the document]
neolitec: Provider setup problem. I didn't found where to setup the provider somewhere else. Come see me for that.

packages/oui-pagination/src/index.spec.js
```diff
@@ -0,0 +1,336 @@
...
+            expect(paginationConfiguration.pageSize).toEqual(50);
```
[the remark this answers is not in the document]
neolitec: ok

packages/oui-pagination/src/index.spec.js
neolitec: Sometime I like to organise a unit test to setup a small scenario. Maybe it's not the purpose of unit tests but we don't have integration tests, now... It could be rewritten as pure unit tests.

buttons[1] is no readable, please, you know that it's the second element of the buttons array described a few lines above ;)

packages/oui-pagination/src/index.spec.js
neolitec: At least, my tests are uniform!

packages/oui-pagination/src/index.spec.js
neolitec: Indeed, it's not PageSizeMax/pageSize...
I don't see another solution than write a comment to explain why 2.

packages/oui-pagination/src/index.spec.js
neolitec: Updated, I hope better.

PR #41, opened 2017-12-12
refactor(oui-datagrid): new paging service

jimmyfortinx:
Demo: http://refactor_paging_service_staging.ui-kit.ovh/#!/oui-angular/datagrid

I will do another pass to remove unused parameters in DatagridController constructor.

### DIFF
--- a/packages/oui-datagrid/src/datagrid.controller.js
+++ b/packages/oui-datagrid/src/datagrid.controller.js
@@ -37,12 +37,12 @@ export default class DatagridController {
             const builtColumns = this.ouiDatagridColumnBuilder.build(columnElements, this.$scope);
             this.columns = builtColumns.columns;
 
-            if (this.rowsLoader || this.rowLoader) {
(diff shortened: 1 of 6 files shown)


packages/oui-datagrid/src/datagrid.controller.js
```diff
@@ -50,11 +36,14 @@ export default class DatagridController {
 
             const builtColumns = this.ouiDatagridColumnBuilder.build(columnElements, this.$scope);
             this.columns = builtColumns.columns;
-            this.currentSorting = builtColumns.currentSorting;
 
-            // Once columns has been processed,
-            // we can proceed with the first loading
-            this.initPage();
+            if (this.rowsLoader || this.rowLoader) {
```
neolitec: `rowLoader` is only used to retrieve a complete single, row. This doesn't systematically use a remote paging.
=> `loadRowsData` should be able to called with a local pagination.

jimmyfortinx: Nice catch, it is now fixed!

PR #46, opened 2017-12-19
test(ouiPagination): refactor pagination tests

neolitec:


packages/oui-pagination/src/index.spec.js
```diff
@@ -302,6 +305,9 @@ describe("ouiPagination", () => {
                     </oui-pagination>
                 `);
 
+                // Page sizes list should not list page size above pageSizeMax (50).
```
tremblaymath: I understand better with that comments thanks!
IMHO I think if you rename your tests, you will not need big comments

maxPageSize = 50
- Should display sizes lower or equal to max page size in page size list
//Page size list = [25, 50, 100, 200]
expect(pageSizesButtons).toEqual([25, maxPageSize]); (it's more meaningfull than length 2)

- Should add max page size value to the page size list if it doesn't exist
//Page size list = [25, 50]
expect(pageSizesButtons).toEqual([25, 50, maxPageSize]);

packages/oui-pagination/src/index.spec.js
```diff
@@ -318,18 +324,36 @@ describe("ouiPagination", () => {
...
+                // Page sizes list should not list page size above pageSizeMax (80).
+                // So, instead of displaying 4 values (see customPageSizesList = [25, 50, 100, 200]) the page size list
+                // is limited to all values lower or equal to pageSizeMax and pageSizeMax: [25, 50, 80] (length: 3).
+                const pageSizesButtons = getPageSizeButtons(element);
 
                 expect(pageSizesButtons.length).toEqual(3);
                 expect(pageSizesButtons[2].innerHTML).toEqual("80");
+            });
+
+            it("should be reinitilized", () => {
```
tremblaymath: When it should be reinitilized? 
should be reinitilized when  pageSizeMax changes?

PR #53, opened 2018-01-03
feat(oui-radio) : allow html in description

virgotopia:


AxelPeter:
I think it was on purpose that we not allow html value on this attribute.

PR #58, opened 2018-01-09
fix(datagrid): handle "data-" prefixes on oui-column

neolitec:


packages/oui-datagrid/src/datagrid-column-builder.service.js
```diff
@@ -87,6 +87,17 @@ export default class DatagridColumnBuilder {
         return {};
     }
 
+    static hasAttribute (element, attributeName) {
+        const attributes = [];
+        Array.from(element.attributes).forEach(attribute => attributes.push(attribute.name));
+        Object.keys(element.dataset).forEach(dataKey => attributes.push(dataKey));
```
jimmyfortinx: Why do you store it ? You don't use it and I don't found it in the current version of the service.

packages/oui-datagrid/src/index.spec.js
```diff
@@ -636,6 +636,40 @@ describe("ouiDatagrid", () => {
...
+                // Check data-sortable and data-title
+                const headerCells = getHeaderRow(element).find("th");
+                expect(headerCells[0].innerHTML).toEqual("First name");
+                expect(headerCells[0].className).toContain("oui-datagrid__header_sortable-asc");
+                expect(headerCells[1].innerHTML).toEqual("Last name");
+                expect(headerCells[1].className).toContain("oui-datagrid__header_sortable");
+
+                // Check data-property
+                const firstRow = getRow(element, 0);
+                expect(getCell(firstRow, 0).text().trim()).toEqual("Ann");
```
jimmyfortinx: `fakeRow[0].firstName` instead of hardcoding `Ann`. It will be easier if you need to change the dataset later.

packages/oui-datagrid/src/index.spec.js
```diff
@@ -636,6 +636,40 @@ describe("ouiDatagrid", () => {
...
+                                <oui-action-menu-item text="Action 2"></oui-action-menu-item>
+                            </oui-action-menu>
+                        </oui-datagrid>
+                    `, {
+                        rows: fakeData.slice(0, 5)
+                    }
+                );
+
+                // Check data-sortable and data-title
+                const headerCells = getHeaderRow(element).find("th");
```
jimmyfortinx: You can do this instead, to correspond to what I already do before (unless I missed something):

```javascript
const $headerRow = getHeaderRow(element);

expect(isSortableCell(getHeaderCell($headerRow, 0))).toBe(true);
expect(getHeaderCell($headerRow, 0).html()).toEqual("First name");
expect(getHeaderCell($headerRow, 0).class()).toContain("oui-datagrid__header_sortable-asc");
```

packages/oui-datagrid/src/datagrid-column-builder.service.js
jimmyfortinx: My bad, I missread the code. Ignore my comment.

packages/oui-datagrid/src/index.spec.js
jimmyfortinx: Saw together, it OK because data is sorted. Ignore my comment.

packages/oui-datagrid/src/index.spec.js
neolitec: done

packages/oui-datagrid/src/index.spec.js
neolitec: Ann is not the first record of the dataset.

PR #61, opened 2018-01-11
fix(datagrid): add scope context to process header titles

neolitec:

### DIFF
--- a/packages/oui-datagrid/src/datagrid.controller.js
+++ b/packages/oui-datagrid/src/datagrid.controller.js
@@ -78,6 +78,22 @@ export default class DatagridController {
 
         this.columns = builtColumns.columns;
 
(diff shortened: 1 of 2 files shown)


packages/oui-datagrid/src/datagrid.controller.js
```diff
@@ -78,6 +78,22 @@ export default class DatagridController {
 
         this.columns = builtColumns.columns;
 
+        this.columns.forEach(column => {
+            if (column.title) {
+                return;
+            }
+
+            column.disableWatcher = this.$scope.$watch(
+                () => this.ouiDatagridColumnBuilder.buildTitle(column.rawTitle, this.getParentScope()),
```
JDownloader: Indent could more clear here, declaring those function before using them could be a way to unclutter

PR #65, opened 2018-01-17
fix(*): fix removeAttr parameter for jqLite

AxelPeter:
Unlike what was done for fixing removeAttr bug from what jQuery documentation said (https://github.com/ovh-ux/ovh-ui-angular/pull/50), jqLite doesn't support multiple attributes: https://docs.angularjs.org/api/ng/function/angular.element

So we have to call `.removeAttr()` for each attributes.

### DIFF
--- a/packages/oui-action-menu/src/action-menu-item.controller.js
+++ b/packages/oui-action-menu/src/action-menu-item.controller.js
@@ -24,19 +24,19 @@ export default class {
     }
 
     $postLink () {
-        this.$timeout(() => {
-            let compiled;
+        let compiled;
 
-            this.$element.removeAttr("aria-label");
+        this.$timeout(() =>
+            this.$element.removeAttr("aria-label")
+        );
 
-            if (this.$attrs.onClick) {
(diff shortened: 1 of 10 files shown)


tremblaymath:
@AxelPeter It's strange, if it did not work, why did the tests pass? Are they well done?

e.g. for back-button
        it("should have and id and name attribute, and remove it from root component"
        it("should have and aria-label attribute, and remove it from root component"

AxelPeter:
@tremblaymath When i begun my tests for the navbar, my tests wouldn't pass for oui-button and oui-back-button. Then when i begun to create a fix for this issue, i saw that the change was already done for these two, but not the others. See the changed files :)

There are only oui-button and oui-back-button that test the remove of the attributes on the root component.

AxelPeter:
Ok, i see that they were changed with this PR: https://github.com/ovh-ux/ovh-ui-angular/pull/59
That i explained why :)

tremblaymath:
👍

PR #73, opened 2018-01-26
feat(checkbox): add required attribute

neolitec:

### DIFF
--- a/packages/oui-checkbox/src/checkbox.controller.js
+++ b/packages/oui-checkbox/src/checkbox.controller.js
@@ -1,3 +1,5 @@
+import { addBooleanParameter } from "@oui-angular/common/component-utils";
+
 export default class {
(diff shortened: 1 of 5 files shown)


packages/oui-checkbox/src/checkbox.controller.js
```diff
@@ -9,6 +11,8 @@ export default class {
     }
 
     $postLink () {
+        addBooleanParameter(this, "required");
```
AxelPeter: Like the disabled attribute, it should be in `$onInit`. And if we use this helper method, we can change the check on disabled with `addBooleanParameter(this, "disabled")` too :)

neolitec: Oops yes, it must be in `$onInit`.

PR #74, opened 2018-01-29
Oui field

neolitec:

### DIFF
--- a/packages/oui-field/src/field.html
+++ b/packages/oui-field/src/field.html
@@ -1,7 +1,7 @@
 <div class="oui-field">
     <label class="oui-field__label oui-label"
         ng-attr-for="{{ ::$ctrl.for }}"
-        ng-if="$ctrl.label"
+        ng-if="::!!$ctrl.label"
         ng-bind="::$ctrl.label"></label>
     <div class="oui-field__control" ng-transclude>
 
@@ -25,7 +25,7 @@
         </span>
     </div>
     <div class="oui-field__helper"
-        ng-if="$ctrl.helpText"
+        ng-if="::!!$ctrl.helpText"
         ng-bind="::$ctrl.helpText">
     </div>
 </div>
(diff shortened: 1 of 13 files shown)


packages/oui-field/src/field.html
```diff
@@ -0,0 +1,31 @@
+<div class="oui-field">
+    <label class="oui-field__label oui-label"
+        ng-attr-for="{{ ::$ctrl.for }}"
+        ng-if="$ctrl.label"
```
tremblaymath: one time binding could be appropriate

packages/oui-field/src/index.spec.js
```diff
@@ -0,0 +1,355 @@
+describe("ouiField", () => {
```
tremblaymath: Nice tests, easy to read !

I think that testing displayed messages could be interesting :
- to validate that error message descriptions settled in provider are used and well formatted with parameters.
- to validate that only one message is displayed with correct priority

packages/oui-field/src/field.html
```diff
@@ -0,0 +1,31 @@
...
+            <span ng-message="maxlength"
+                ng-bind="$ctrl.getErrorMessage('maxlength')"></span>
+            <span ng-message="min"
+                ng-bind="$ctrl.getErrorMessage('min')"></span>
+            <span ng-message="max"
+                ng-bind="$ctrl.getErrorMessage('max')"></span>
+        </span>
+    </div>
+    <div class="oui-field__helper"
+        ng-if="$ctrl.helpText"
```
tremblaymath: one time binding could be appropriate

packages/oui-field/src/field.html
neolitec: One time binding has no effet with ng-if

packages/oui-field/src/field.html
neolitec: +1

packages/oui-field/src/index.spec.js
neolitec: Yes, + parameters

PR #75, opened 2018-01-30
feat(modal): add new component for modal

neolitec:
https://github.com/ovh-ux/ovh-ui-kit/pull/137


packages/oui-modal/src/modal.controller.js
```diff
@@ -0,0 +1,2 @@
+export default class ModalController {
+}
```
AxelPeter: Empty controller, is it necessary in a separate file ?

neolitec: I created it without thinking about it until I realized that it was useless :D ... but I forgot to remove it :/

PR #76, opened 2018-01-31
fix(datagrid): prevent useless calls to row-loader

neolitec:


packages/oui-datagrid/src/paging/datagrid-paging.service.js
```diff
@@ -2,10 +2,11 @@ import DatagridLocalPaging from "./datagrid-local-paging";
 import DatagridRemotePaging from "./datagrid-remote-paging";
 
 export default class {
-    constructor ($q, orderByFilter) {
+    constructor ($q, $timeout, orderByFilter) {
         "ngInject";
 
         this.$q = $q;
+        this.$timeout = $timeout;
```
AxelPeter: Set, but not used in the Class

neolitec: It's used in `DatagridLocalPaging`.

PR #83, opened 2018-02-05
chore(): fix eslint errors

neolitec:


JDownloader:
I would personally remove/modify current eslint rule causing this, code was more readable before

PR #85, opened 2018-02-07
feat(back-button): Add title attribute

tremblaymath:


packages/oui-back-button/src/back-button.html
```diff
@@ -1,8 +1,11 @@
-<button type="button"
-        class="oui-button oui-button_link oui-button_icon-left"
-        ng-attr-aria-label="{{:: $ctrl.ariaLabel }}"
-        ng-attr-id="{{:: $ctrl.id }}"
-        ng-attr-name="{{:: $ctrl.name }}"
-        ng-click="$ctrl.onBtnClick()">
-  <i class="oui-icon oui-icon-chevron-left" aria-hidden="true"></i>
-</button>
+<div class="oui-back-button">
```
AxelPeter: I think we should set the classname `oui-back-button` on the root element `<oui-back-button>`, to avoid a useless `<div>`.

tremblaymath: Yes done!

PR #89, opened 2018-02-13
feat(textarea): add new component

neolitec:

### DIFF
--- a/packages/oui-textarea/src/index.spec.js
+++ b/packages/oui-textarea/src/index.spec.js
@@ -1,0 +1,257 @@
+describe("ouiTextarea", () => {
+    let $sniffer;
+    let $timeout;
+    let TestUtils;
+
+    const getRoot = elm => elm[0].querySelector(".oui-textarea");
+    const getTextarea = elm => elm[0].querySelector("textarea");
+    const getFooter = elm => elm[0].querySelector(".oui-textarea__footer");
+
+    const disabledClass = "oui-textarea_disabled";
+    const readonlyClass = "oui-textarea_readonly";
+    const focusClass = "oui-textarea_active";
+
+    beforeEach(angular.mock.module("oui.textarea"));
+    beforeEach(angular.mock.module("oui.test-utils"));
+
+    beforeEach(inject((_$sniffer_, _$timeout_, _TestUtils_) => {
+        $sniffer = _$sniffer_;
+        $timeout = _$timeout_;
+        TestUtils = _TestUtils_;
+    }));
+
+    describe("Component", () => {
+        it("should display a textarea", () => {
+            const id = "text-1";
+            const name = "text-name-1";
+
+            const element = TestUtils.compileTemplate(`
+                <oui-textarea
+                    id="${id}"
+                    name="${name}"></oui-textarea>
+            `);
+
+            const textarea = getTextarea(element);
+
+            expect(textarea).toBeDefined();
+            expect(textarea.getAttribute("id")).toEqual(id);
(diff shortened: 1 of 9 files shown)


packages/oui-textarea/src/textarea.html
```diff
@@ -0,0 +1,22 @@
+<div class="oui-textarea"
```
AxelPeter: We can avoid this useless DIV by adding all those classes on the root element, to generate only:
```
<oui-textarea class="oui-textarea">
    <textarea></textarea>
</oui-textarea>
```
The `disabled` and `readonly` part could be changed with $onChange if necessary

packages/oui-textarea/src/textarea.html
```diff
@@ -0,0 +1,22 @@
...
+    ng-class="{
+        'oui-textarea_disabled': $ctrl.disabled,
+        'oui-textarea_readonly': $ctrl.readonly
+    }">
+    <textarea class="oui-textarea__textarea"
```
AxelPeter: This classname is... something :)
I know it's not the tagname but maybe we could go with `oui-textarea__input` or `oui-textarea__body`

jimmyfortinx:
(optional) If I click on the footer the focus is not moved to the textarea. Also, if the focus was already to the textarea and I click on the footer the focus is changed to something else so I can't continue enter text. It is not really a critical bug, so it can be fixed later.

packages/oui-textarea/src/index.spec.js
```diff
@@ -0,0 +1,257 @@
...
+            it("should be disabled without provider value", () => {
+                const element = TestUtils.compileTemplate(`
+                    <oui-textarea disabled></oui-textarea>
+                `);
+
+                const $textarea = angular.element(getTextarea(element));
+                expect($textarea.prop("disabled")).toBeTruthy();
+            });
+
+            it("should show be disabled with the binding set to true", () => {
```
jimmyfortinx: typo: `should be disabled...` instead of `should show...` ?

packages/oui-textarea/src/index.spec.js
```diff
@@ -0,0 +1,257 @@
...
+                const element = TestUtils.compileTemplate(`
+                    <oui-textarea disabled="true"></oui-textarea>
+                `);
+
+                const $textarea = angular.element(getTextarea(element));
+                expect($textarea.prop("disabled")).toBeTruthy();
+            });
+
+            it("should show not be disabled with the binding set to false", () => {
```
jimmyfortinx: typo: `should not be disabled...` instead of `should show...` ?

packages/oui-textarea/src/textarea.html
neolitec: Something I know. I thought about `input`, but it wasn't an input. I also thought about `body`  but it's not really a `body` ... `textarea` is a form controller... It could be `control`, I think.

packages/oui-textarea/src/textarea.html
neolitec: So bloated code. I really don't like this old broken deprecated UI framework...

neolitec: I also thought about it and when I come to wonder if I should use the component tag... I try to know if this could could be used with `ng-repeat`. If no, I don't make the effort.

packages/oui-textarea/src/textarea.html
AxelPeter: Yes, I find `oui-textarea__control` better too :)

PR #97, opened 2018-02-21
Oui field select

neolitec:


packages/oui-field/src/index.spec.js
```diff
@@ -241,6 +242,9 @@ describe("ouiField", () => {
                 $control.triggerHandler("focus");
                 expect(getError(element)).toBeNull();
             });
+
+            // TODO
```
tremblaymath: Todo

PR #100, opened 2018-02-26
feat(select-picker): create component

rfougier:


packages/oui-select-picker/src/select-picker.controller.js
```diff
@@ -0,0 +1,55 @@
...
+        this.$transclude = $transclude;
+    }
+
+    $postLink () {
+        // Sometimes the digest cycle is done before dom manipulation,
+        // So we use $timeout to force the $apply
+        this.$timeout(() => {
+            this.$element
+                .removeAttr("id")
+                .removeAttr("name");
```
whispyy: (optional) may be removeAttribute concerning aria-label ?

AxelPeter:
You should remove the files from the build

PR #102, opened 2018-02-26
feat(criteria-container): add new component

neolitec:
**Merge this AFTER https://github.com/ovh-ux/ovh-ui-angular/pull/105.**

Only the criteria container is in this pull request. Not any link to datagrid here.

### DIFF
--- a/packages/oui-criteria-container/src/criteria-container.component.js
+++ b/packages/oui-criteria-container/src/criteria-container.component.js
@@ -1,0 +1,11 @@
+import controller from "./criteria-container.controller";
+
+export default {
+    template: "<ng-transclude></ng-transclude>",
+    transclude: true,
+    controller,
+    bindings: {
+        criteria: "=model",
+        onChange: "&?"
+    }
+};
--- a/packages/oui-criteria-container/src/criteria-container.component.js
+++ b/packages/oui-criteria-container/src/criteria-container.component.js
@@ -5,7 +5,6 @@ export default {
     transclude: true,
     controller,
     bindings: {
-        criteria: "=model",
-        onChange: "&?"
+        onChange: "&"
     }
 };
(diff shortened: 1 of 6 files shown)


packages/oui-criteria-container/src/index.spec.js
```diff
@@ -0,0 +1,108 @@
...
+
+        it("should set all criterias", () => {
+            const criteria = [criterion, criterion2];
+            expect(controller.criteria.length).toEqual(0);
+
+            controller.set(criteria);
+            expect(controller.criteria.length).toEqual(2);
+        });
+
+        it("should delete all criterias", () => {
```
tremblaymath: Criteria without "s" to change everywhere

packages/oui-criteria-container/src/criteria-container.component.js
```diff
@@ -0,0 +1,11 @@
+import controller from "./criteria-container.controller";
+
+export default {
+    template: "<ng-transclude></ng-transclude>",
+    transclude: true,
+    controller,
+    bindings: {
+        criteria: "=model",
+        onChange: "&?"
```
tremblaymath: You should remove "?" else you'll be forced to check if (this.onChange) before triggering on
Change.
To reproduce the bug, remove the callback for on-change and trigger onChange event. An exception will raise.

packages/oui-criteria-container/src/criteria-container.component.js
```diff
@@ -0,0 +1,11 @@
+import controller from "./criteria-container.controller";
+
+export default {
+    template: "<ng-transclude></ng-transclude>",
+    transclude: true,
+    controller,
+    bindings: {
+        criteria: "=model",
```
tremblaymath: As discussed, I suggest to remove the two-way to respect the initial decision "Use two-way bindings with ng-model and one-way bindings for read-only components"

packages/oui-criteria-container/src/index.spec.js
```diff
@@ -1,6 +1,6 @@
 import CriteriaContainerController from "./criteria-container.controller";
 
-describe("ouiCriteriaContainer", () => {
+fdescribe("ouiCriteriaContainer", () => {
```
tremblaymath: fdescribe to remove

PR #107, opened 2018-02-27
fix(oui-modal): README error on-dismiss doc

whispyy:
on the API part README.md : on-close was wrote instead of on-dismiss


packages/oui-modal/README.md
```diff
@@ -47,4 +47,4 @@
 | `primary-label`     | string   | @?      | yes              |                        |                     | confirmation label                               |
 | `secondary-action`  | function | &?      |                  |                        |                     | cancellation callback                            |
 | `secondary-label`   | string   | @?      | yes              |                        |                     | cancellation label                               |
-| `on-close`          | function | &?      |                  |                        |                     | close callback                                   |
+| `on-dismiss`        | function | &?      |                  |                        |                     | close callback                                   |
```
jimmyfortinx: to keep the same vocabulary you should change the "close callback" sentence below too

whispyy: ok

PR #109, opened 2018-02-28
feat(field): add custom error message and ng-pattern support

JDownloader:


packages/oui-field/README.md
```diff
@@ -149,7 +160,20 @@
...
-| `label`           | function        | @?      | yes              |                           |                     | a text to help fill the form field                |
+| Attribute         | Type            | Binding | One-time binding | Values      | Default             | Description                                       |
+| ----              | ----            | ----    | ----             | ----        | ----                | ----                                              |
+| `error-messages`  | object          | <?      | no               |             |                     | a dictionary to override default messages         |
+| `helper-text`     | string          | @?      | yes              |             |                     | the field label                                   |
+| `label`           | function        | @?      | yes              |             |                     | a text to help fill the form field                |
+
+### error-messages
+
+| Attribute         | Type            | Binding | One-time binding | Values      | Default                                   | Description                 |
```
neolitec: It's not really a tag attribute, so... some columns should not be there like "Binding", "One-time binding"...
Maybe it's could be sufficient to only show the default object and explaining that it can be overridden.

packages/oui-field/src/field.controller.js
```diff
@@ -22,7 +22,8 @@ const VALIDATION_PARAMETERS = {
     min: ["min", "ng-min"],
     max: ["max", "ng-max"],
     minlength: ["minlength", "ng-minlength"],
-    maxlength: ["maxlength", "ng-maxlength"]
+    maxlength: ["maxlength", "ng-maxlength"],
+    custumValidation: ["pattern", "ng-pattern"]
```
neolitec: *custum* -> *custom*
Or could it be simply named "pattern"?

packages/oui-field/src/index.spec.js
```diff
@@ -466,6 +520,40 @@ describe("ouiField", () => {
             });
         });
 
+        describe("with validation", () => {
+            it("should retrieve custom error messages", () => {
+                const message = "Username must be a least 6 characters.";
```
neolitec: (optional but should not be) I don't remember if it has been done before but I think you should also test that interpolation works in custom messages.

packages/oui-field/README.md
```diff
@@ -30,6 +30,17 @@
...
+
+<oui-field label="{{'Username'}}"
+    error-messages="{pattern: 'Username must be alphabetic with a length between 3 and 8.'}">
+    <input
+        class="oui-input"
+        type="text"
+        id="username"
+        name="username"
+        ng-model="$ctrl.user.username"
+        ng-pattern="/^[a-z]{3,8}$/">
```
neolitec: (optional) Maybe `[a-zA-Z]` to exactly match with the error message.

packages/oui-field/README.md
tremblaymath: I think it could be replaced with a configuration section like for pagination component.
An exemple of provider configuration with default message could be interesting

packages/oui-field/src/field.controller.js
tremblaymath: +1 for pattern

packages/oui-field/src/index.spec.js
```diff
@@ -466,6 +520,40 @@ describe("ouiField", () => {
             });
         });
 
+        describe("with validation", () => {
```
tremblaymath: Good test for custom message!
We should test default error messages too. Almost identical to this one

packages/oui-field/README.md
```diff
@@ -30,6 +30,17 @@
         name="email"
         ng-model="$ctrl.user.email">
 </oui-field>
+
+<oui-field label="{{'Username'}}"
+    error-messages="{pattern: 'Username must be alphabetic with a length between 3 and 8.'}">
```
tremblaymath: I would like to see parameters in this example

packages/oui-field/README.md
JDownloader: Yeah, the binding is for the object containing the messages, I was was not sure on how to present it. I will take a look at the pagination section to try to improve that.

packages/oui-field/src/index.spec.js
JDownloader: This could be done in a new test, only for testing the interpolation.

packages/oui-field/README.md
```diff
@@ -149,7 +160,20 @@
 
 ### oui-datagrid
```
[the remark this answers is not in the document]
JDownloader: @tremblaymath I'm not what you want about that subtitle

packages/oui-field/src/index.spec.js
JDownloader: Yeah, but it the test should not enforce a string but verify it's actually using the one from the provider

packages/oui-field/README.md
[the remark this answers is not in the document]
tremblaymath: oui-datagrid has no sense in field component documentation.
To remove

packages/oui-field/README.md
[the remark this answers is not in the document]
neolitec: An helper text or an error, not both of them...
Maybe something like "This input is valid only if..." for helper text.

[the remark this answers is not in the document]
JDownloader: I would avoid also doing both, I think the error is easy to trigger like the lastname field.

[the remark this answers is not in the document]
tremblaymath: easy to trigger yes!
But the user doesn't know the validations without clicking on "Show the example" and reading the code.
"I think the error is easy to trigger like the lastname field." There is already a help text for lastname "At least 3 chars" and it's a good thing that we should reproduce

packages/oui-field/src/index.spec.js
tremblaymath: Yes I agree, provider should be mocked and we should only "verify it's actually using the one from the provider"
There was already missing tests before you start

PR #110, opened 2018-02-28
fix(select): make match optional (handle string list)

whispyy:
In case of string list instead of objects list passing to your item. You won't be able to match object attribute. So, to handle this case match attribute has been turned to optional.

http://angular_fix_oui_select.ui-kit.ovh/#!/oui-angular/select

### DIFF
--- a/packages/oui-select/README.md
+++ b/packages/oui-select/README.md
@@ -4,7 +4,7 @@
 
 ## Usage
 
-### Basic
+### Basic (Object array)
 
 ```html:preview
 <oui-select name="country"
@@ -16,6 +16,18 @@
     match="name"
     data-align="start">
     <span ng-bind="$item.name"></span>
+</oui-select>
+```
+
+### Basic (String array)
+
+```html:preview
+<oui-select name="country"
+    model="$ctrl.country"
+    items="['a', 'b', 'c']"
+    required
+    data-align="start">
+    <span ng-bind="$item"></span>
 </oui-select>
 ```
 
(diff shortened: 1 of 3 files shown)


packages/oui-select/README.md
```diff
@@ -19,6 +19,18 @@
 </oui-select>
 ```
 
+```html:preview
```
tremblaymath: Add a title to differentiate it eg: string array vs object array

packages/oui-select/src/index.spec.js
```diff
@@ -49,6 +49,28 @@ describe("ouiSelect", () => {
...
+                    data-title="${title}"
+                    placeholder="${placeholder}"
+                    items="$ctrl.countries.name"
+                    data-align="start">
+                    <span ng-bind="$item"></span>
+                </oui-select>`, {
+                countries: data
+            });
+
+            expect(angular.element(getContainer(element)).attr("title")).toEqual(title);
```
tremblaymath: You should test that the select is populated with item values

packages/oui-select/src/index.spec.js
```diff
@@ -49,6 +49,23 @@ describe("ouiSelect", () => {
             expect(getDropdown(element)).toBeDefined();
         });
 
+        it("should display the select component without passing match attribute", () => {
```
tremblaymath: The test should be "should display all the choices (stringArray)" and assert on choices.
This line isn't tested "matchElement.html("{{$select.selected}}");".

packages/oui-select/README.md
```diff
@@ -19,6 +19,18 @@
 </oui-select>
 ```
 
+### Basic (String array)
+
+```html:preview
+<oui-select name="country"
+    model="$ctrl.country"
+    items="['a', 'b', 'c']"
```
tremblaymath: 👍

packages/oui-select/src/index.spec.js
tremblaymath: NVM I saw the new test should display all the choices (objectArray)
This one isn't useful now

PR #117, opened 2018-03-05
feat(oui-radio): add transclude and thumbnail style options

rfougier:


packages/oui-radio/src/index.spec.js
```diff
@@ -113,6 +116,40 @@ describe("ouiRadio", () => {
             });
         });
 
+        describe("thumbnail attribute", () => {
+            it("should display a classic radio when no attribute", () => {
+                const element = TestUtils.compileTemplate("<oui-radio></oui-radio>");
+
+                const mainBlockElement = getMainBlockElement(element);
+                expect(angular.element(mainBlockElement).hasClass("oui-radio_thumbnail")).toBe(false);
```
JDownloader: the absence of a class is maybe not sufficient to prove the element is displayed

neolitec: This is not an end to end test so it's correct for me.

PR #118, opened 2018-03-05
fix(oui-select): fix readme controller selector

whispyy:
Fix examples (bad controllers declared).


neolitec:
Already done and merged ^^

PR #120, opened 2018-03-07
fix(*): button type explicit

neolitec:


packages/oui-modal/src/modal.html
```diff
@@ -16,12 +16,12 @@ <h2 class="oui-modal__title" ng-bind="::$ctrl.title"></h2>
...
       ng-click="$ctrl.secondaryAction()">
-    </button><button ng-if="$ctrl.primaryLabel"
+    </button>
+    <button ng-if="$ctrl.primaryLabel"
       class="oui-button oui-button_primary"
-      type="button"
+      type="submit"
       ng-bind="::$ctrl.primaryLabel"
       ng-click="$ctrl.primaryAction()">
-      Save
```
JDownloader: By default it could be ok and cancel

PR #123, opened 2018-03-09
feat(modal): add overlay and spinner to modal attribute while loading

whispyy:


packages/oui-modal/README.md
```diff
@@ -8,16 +8,16 @@
 <div ng-init="$ctrl.cancel = false; $ctrl.confirm = false"
     class="oui-doc-preview-only-keep-children">
 <oui-modal
-    data-title="Modal title"
+    title="Modal title"
```
neolitec: `data-` is needed here! `title` generates a native tooltip...

packages/oui-modal/src/modal.html
```diff
@@ -4,24 +4,34 @@
...
+    </button>
   </div>
   <div class="oui-modal__body oui-modal__body_no-icon">
-    <h2 class="oui-modal__title" ng-bind="::$ctrl.title"></h2>
+    <div class="oui-modal__loader" ng-if="$ctrl.loading">
+      <oui-spinner></oui-spinner>
+    </div>
+    <h2 class="oui-modal__title"
+        ng-bind="::$ctrl.title"
+        ng-if="$ctrl.title">
```
AxelPeter: Since `$ctrl.title` is one-time binded, you can go with `ng-if="::!!$ctrl.title"`. 
AngularJS keep a watcher if the value is `undefined`.

packages/oui-modal/README.md
tremblaymath: done

packages/oui-modal/src/modal.html
tremblaymath: done

PR #125, opened 2018-03-09
Feat/link filters to datagrid

neolitec:
UU-747


neolitec:
Merged with https://github.com/ovh-ux/ovh-ui-angular/pull/126

PR #126, opened 2018-03-12
Feat/datagrid top bar

neolitec:
**Merge this AFTER https://github.com/ovh-ux/ovh-ui-angular/pull/125.**

UU-777

Staging: https://ovh_ui_kit_documentation-feat_datagrid_top_bar_staging.uxci.ovh/#!/oui-angular/datagrid

### DIFF
--- a/packages/oui-chips/src/chips.controller.js
+++ b/packages/oui-chips/src/chips.controller.js
@@ -1,5 +1,4 @@
 import { addBooleanParameter } from "@oui-angular/common/component-utils";
-import { cloneDeep } from "lodash";
 
 export default class {
     constructor ($attrs, $element, $timeout) {
@@ -14,7 +13,7 @@ export default class {
         addBooleanParameter(this, "closable");
         addBooleanParameter(this, "stacked");
 
-        this.items = this.items ? cloneDeep(this.items) : [];
+        this.items = this.items ? angular.copy(this.items) : [];
     }
 
     $postLink () {
--- a/packages/oui-chips/src/chips.controller.js
+++ b/packages/oui-chips/src/chips.controller.js
@@ -1,4 +1,5 @@
 import { addBooleanParameter } from "@oui-angular/common/component-utils";
+import { cloneDeep } from "lodash";
 
 export default class {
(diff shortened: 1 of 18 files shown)


packages/oui-criteria-container/src/criteria-container.controller.js
```diff
@@ -39,6 +39,16 @@ export default class CriteriaController {
     }
 
     add (criterion) {
+        // Delete same preview criterion if it exists.
+        const previewCriterion = cloneDeep(criterion);
```
mlarose: at a quick glance, I find that there is a lot of preventive/defensive copying using cloneDeep. 
eventually this could get expensive performance wise. perhaps we should consider immutablejs.

neolitec: In fact I think cloneDeep isn't necessary... It should work with a simple clone.
The performance is not really impacted because we'll never have a filter with a huge number of criteria.
ImmutableJS could be interesting for more specific usage but also at a cost of a significant increasing of dist size unfortunately.

packages/oui-chips/src/chips.controller.js
```diff
@@ -12,6 +13,8 @@ export default class {
     $onInit () {
         addBooleanParameter(this, "closable");
         addBooleanParameter(this, "stacked");
+
+        this.items = this.items ? cloneDeep(this.items) : [];
```
AxelPeter: Why `cloneDeep` ? `angular.clone` should be enough, it does a deep clone.

packages/oui-chips/src/index.spec.js
```diff
@@ -61,7 +63,29 @@ describe("ouiChips", () => {
...
+                            items="$ctrl.items">
+                        </oui-chips>
+                    </oui-criteria-container>
+                `, {
+                    items: mockData.criteria,
+                    onChangeSpy
+                });
+
+                const criteriaContainerController = component.controller("ouiCriteriaContainer");
+                criteriaContainerController.criteria = cloneDeep(mockData.criteria);
```
AxelPeter: same here :)

PR #131, opened 2018-03-15
fix(oui-field): support validations prefixed with data-ng

tremblaymath:
It's a ugly fix, we should change the pattern to not be forced to hardcode all possibilities. To discuss after...

### DIFF
--- a/packages/oui-field/src/field.controller.js
+++ b/packages/oui-field/src/field.controller.js
@@ -13,11 +13,11 @@ const CONTROLS_SELECTORS = [
 ];
 
 const VALIDATION_PARAMETERS = {
-    min: ["min", "ng-min"],
-    max: ["max", "ng-max"],
-    minlength: ["minlength", "ng-minlength"],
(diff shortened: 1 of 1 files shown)


packages/oui-field/src/field.controller.js
```diff
@@ -13,11 +13,11 @@ const CONTROLS_SELECTORS = [
 ];
 
 const VALIDATION_PARAMETERS = {
-    min: ["min", "ng-min"],
-    max: ["max", "ng-max"],
-    minlength: ["minlength", "ng-minlength"],
-    maxlength: ["maxlength", "ng-maxlength"],
-    pattern: ["pattern", "ng-pattern"]
+    min: ["min", "ng-min", "ngMin"],
```
neolitec: We also could do something like `.map(attribute => _.camelCase(attribute))` in _component-utils.js_.

tremblaymath: to discuss after

PR #138, opened 2018-03-20
feat(oui-back-button): add the possibility to handle href

whispyy:
UU-797

### DIFF
--- a/packages/oui-back-button/src/back-button.html
+++ b/packages/oui-back-button/src/back-button.html
@@ -1,10 +1,15 @@
-<button type="button"
+<button ng-if="::!$ctrl.href"
+        type="button"
         class="oui-button oui-button_icon-alone"
         ng-attr-aria-label="{{:: $ctrl.ariaLabel }}"
         ng-attr-id="{{:: $ctrl.id }}"
         ng-attr-name="{{:: $ctrl.name }}"
-        ng-click="$ctrl.onBtnClick()"
-        type="button">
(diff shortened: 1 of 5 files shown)


packages/oui-back-button/src/back-button.html
```diff
@@ -1,10 +1,15 @@
...
         class="oui-button oui-button_icon-alone"
         ng-attr-aria-label="{{:: $ctrl.ariaLabel }}"
         ng-attr-id="{{:: $ctrl.id }}"
         ng-attr-name="{{:: $ctrl.name }}"
-        ng-click="$ctrl.onBtnClick()"
-        type="button">
+        ng-click="$ctrl.onBtnClick()">
   <i class="oui-icon oui-icon-chevron-left" aria-hidden="true"></i>
 </button>
+<a ng-if="$ctrl.href"
```
AxelPeter: We should go with `ng-if="::!!$ctrl.href"` if it's one-time binded

packages/oui-back-button/src/back-button.html
```diff
@@ -1,10 +1,15 @@
-<button type="button"
+<button ng-if="!$ctrl.href"
```
AxelPeter: `ng-if="::!$ctrl.href"` for one-time binding

packages/oui-back-button/src/back-button.html
whispyy: Yes u'right!

PR #139, opened 2018-03-20
feat(oui-form-actions): handle href for cancel button

whispyy:
UU-803


tremblaymath:
I removed on-cancel from default example, the good example is with href and I separated Custom naming example from Event example

PR #143, opened 2018-03-26
feat(oui-form-actions): hide cancel button and don't disable submit

JDownloader:
Some forms don't a cancel boutton so it must be hidden. Guidelines
say that submit button is always enabled, so I've remove the
disabled argument.

Breaks oui-form-actions:disabled, submit function will be called
event if oui-form-actions have a disabled argument.

### DIFF
--- a/packages/oui-form-actions/README.md
+++ b/packages/oui-form-actions/README.md
@@ -17,12 +17,12 @@
 ### Custom naming
 
 ```html:preview
-<div ng-init="$ctrl.disabled = false" class="oui-doc-preview-only-keep-children">
+<div class="oui-doc-preview-only-keep-children">
     <oui-form-actions
       on-submit="$ctrl.submit()"
       href="#"
-      submit-text="Disable"
-      cancel-text="Reset">
+      submit-text="Apply"
+      cancel-text="Close">
     </oui-form-actions>
 </div>
 ```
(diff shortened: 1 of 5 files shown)


packages/oui-form-actions/README.md
```diff
@@ -17,12 +17,12 @@
 ### Custom naming
 
 ```html:preview
-<div ng-init="$ctrl.disabled = false" class="oui-doc-preview-only-keep-children">
+<div class="oui-doc-preview-only-keep-children">
     <oui-form-actions
       on-submit="$ctrl.submit()"
       href="#"
-      submit-text="Disable"
-      cancel-text="Reset">
```
JDownloader: This was done to avoid confusion with button of type ```reset```

packages/oui-form-actions/README.md
```diff
@@ -33,13 +33,21 @@
 ```html:preview
 <div ng-init="$ctrl.disabled = false" class="oui-doc-preview-only-keep-children">
     <oui-form-actions
-      on-submit="$ctrl.disabled = true"
-      on-cancel="$ctrl.disabled = false"
-      disabled="$ctrl.disabled">
```
JDownloader: submit cannot be disabled

whispyy:
I've approved the PR BUT I disapproved CX decision about : removing the possibility to disable a submit button, it's a nonsense.

PR #159, opened 2018-04-16
feat(oui-navbar): replace template by new notifications

JDownloader:


packages/oui-navbar/src/navbar.html
```diff
@@ -104,8 +104,19 @@ <h3 class="oui-navbar-menu__title" ng-bind="::asideLink.headerTitle || asideLink
...
                     ng-click="$ctrl.toggleMenu()"
                     ng-href="{{::subLink.url}}"
-                    ng-bind-html="::subLink.template"
                     oui-navbar-group="{{::asideLink.name}}">
+                    <span class="notification__icon">
+                        <i class="oui-icon oui-icon_bicolor"
+                           ng-class=":: 'oui-icon-' + subLink.level + '_circle'"
+                           aria-hidden="true"></i>
+                    </span>
+                    <div>
```
AxelPeter: I think you should add a classname like `notification__content`, to avoid having a nude `div` in our template

packages/oui-navbar/src/navbar.html
```diff
@@ -104,8 +104,19 @@ <h3 class="oui-navbar-menu__title" ng-bind="::asideLink.headerTitle || asideLink
...
+                        <div>
```
AxelPeter: And here, something like `notification__header`

packages/oui-navbar/src/index.spec.data.json
```diff
@@ -45,20 +45,35 @@
...
+                "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. In fermentum faucibus ante, sed ullamcorper quam pretium ut. Donec et eros at nisi egestas scelerisque quis id erat. Cras suscipit diam id lorem congue aliquam. Integer ornare urna vel turpis blandit, mattis placerat est dictum. Etiam porta venenatis odio tristique condimentum."
+              }, {
+                "subject": "Duis placerat lacus at nibh fringilla",
+                "time": "last week",
+                "level": "success",
+                "url": "#/",
+                "description": "Duis placerat lacus at nibh fringilla, at ultrices nunc suscipit."
+              }, {
+                 "subject": "Mauris ipsum",
+                 "time": "last week",
```
tremblaymath: "last week" isn't a "time", it's a formatted date

I think we should give a date object to the component and the component is responsible to format it in a standardized way. You could create a custom dateFilter or something like that..

packages/oui-navbar/src/index.spec.data.json
```diff
@@ -45,20 +45,35 @@
...
-                "template": "Vestibulum malesuada justo ac lectus tincidunt, <strong>sit amet fringilla lectus iaculis</strong>."
-              },
-              {
-                "title": "Duis placerat lacus at nibh fringilla",
+                "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. In fermentum faucibus ante, sed ullamcorper quam pretium ut. Donec et eros at nisi egestas scelerisque quis id erat. Cras suscipit diam id lorem congue aliquam. Integer ornare urna vel turpis blandit, mattis placerat est dictum. Etiam porta venenatis odio tristique condimentum."
+              }, {
+                "subject": "Duis placerat lacus at nibh fringilla",
+                "time": "last week",
+                "level": "success",
+                "url": "#/",
```
tremblaymath: We should be able to handle onClick to update notification status in the api

tremblaymath:
You should offer a feature to handle click on notification menu in the navbar to aknowledge all notifications

packages/oui-navbar/src/navbar.html
```diff
@@ -99,13 +99,24 @@ <h3 class="oui-navbar-menu__title" ng-bind="::asideLink.headerTitle || asideLink
...
                     oui-navbar-group="{{::asideLink.name}}">
+                    <span class="oui-navbar-notification__icon">
+                        <i class="oui-icon oui-icon_bicolor"
+                           ng-class=":: 'oui-icon-' + subLink.level + '_circle'"
+                           aria-hidden="true"></i>
+                    </span>
+                    <div class="oui-navbar-notification__body">
+                        <div class="oui-navbar-notification__header">
+                            <span class="oui-navbar-notification__title" ng-bind=":: subLink.subject"></span>
+                            <span class="oui-navbar-notification__date" ng-bind="subLink.time"></span>
```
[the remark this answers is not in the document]
JDownloader: Not if I wan't to update it.
1 minute ago
1 hour age
2 hours ago
...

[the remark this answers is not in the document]
JDownloader: maybe time should replace date, see further comments...

packages/oui-navbar/src/index.spec.data.json
JDownloader: moment-angular is not included in the project and I don't think it would be a good idea to do it. User should have control on how date is displayed. "last week" isn't a date either, should I use ```formated_time```?

packages/oui-navbar/src/navbar.html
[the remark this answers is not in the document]
tremblaymath: ok it's not a cx requirement for now but I understand the feature

packages/oui-navbar/src/index.spec.data.json
tremblaymath: Why moment-angular is a blocker?
You could pass a timestamp or something like that and use native js Date.
IMHO The logic to format it should be at one place, in the component, to be standardized and not duplicated everywhere.

packages/oui-navbar/src/navbar.html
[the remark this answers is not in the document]
JDownloader: It less complicated to bind whit time right now, Simon thinks it's a very good idea. I would be so simple to implement...

packages/oui-navbar/src/navbar.html
```diff
@@ -74,21 +74,21 @@
...
+                    ng-if="!!asideLink.subLinks && (asideLink.subLinks | filter: {isActive: 'true', acknowledged: '!true'}).length"
+                    ng-bind="(asideLink.subLinks | filter: {isActive: 'true', acknowledged: '!true'}).length">
                 </span>
             </span>
             <span ng-bind="::asideLink.title"></span>
         </button>
 
         <!-- Notifications Menu -->
-        <div ng-switch-when="notifications"
+        <ul ng-switch-when="notifications"
```
AxelPeter: This should stay a DIV tag (no LI children)

packages/oui-navbar/src/navbar.html
```diff
@@ -74,21 +74,21 @@
             ng-attr-aria-label="{{::!!asideLink.label ? asideLink.label : null}}"
             ng-attr-title="{{::!!asideLink.label ? asideLink.label : null}}"
             ng-class="::!!asideLink.iconClass ? 'oui-navbar-link_icon' : null"
-            ng-click="$ctrl.toggleMenu(asideLink.name)"
+            ng-click="$ctrl.toggleMenu(asideLink.name); asideLink.onClick();"
```
AxelPeter: No check for `onClick()` if undefined ? 

What do you think about adding a parameter for the toggleMenu method `toggleMenu (state, isInternalNav, callback)` and check if is function ?

JDownloader: Is it a bit weird to call ```$ctrl.toggleMenu(asideLink.name, false, asideLink.onClick())```
```toggleMenu``` is already used at multiple place in navbar.html

Comment (AxelPeter) [reply packages/oui-navbar/src/navbar.html]: In fact, I had more in mind
```
$ctrl.toggleMenu(asideLink.name, false, asideLink.onClick)
```
and in the method
```
if (angular.isFunction(callback)) {
   callback();
}
```
This way, it will not disturb the others

JDownloader: Le problème est que ```asideLink.onClick``` prend en argument ```asideLink``` qui n'est pas visible dans le controlleur

PR #164, opened 2018-04-18
Feature/dual list

radireddy:
Added Dual List component, test cases and documentation

Staging and other info in below jira issue number
UU-831


packages/oui-dual-list/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+<component-status cx-design="complete" ux="complete"></component-status>
+
+## Usage
+
+### Basic (Empty component)
+
+```html:preview
+<oui-dual-list></oui-dual-list>
+<div style="margin-bottom: 20px;"></div>
```
AxelPeter: Careful with this kind of inline style. It will be present in the code preview, for the copy/paste.
We have helper classes for these use cases (see `oui-doc-preview-only` and `oui-doc-preview-only-keep-children`).

But I think we could set the margin directly to `.oui-dual-list`, since it's a block component.

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+    constructor ($q, $element, ouiDualListProvider) {
+        "ngInject";
+
+        this.$q = $q;
+        this.$element = $element;
+        this.dualListProvider = ouiDualListProvider;
+    }
+
+    $onInit () {
+        this.sourceListLabel = this.sourceListLabel || this.dualListProvider.translations.sourceListLabel;
```
AxelPeter: We should use `addDefaultParameter()`.
We have some component utils for this use cases, in `component-utils.js`. :)

You'll need $attrs in your controller for this

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+        this.loadingMap = {};
+        this.targetCloseIconStyle = {};
+        this.targetOpenIconStyle = {};
+        this.targetContentStyle = {};
+        this.sourceCloseIconStyle = {};
+        this.sourceOpenIconStyle = {};
+        this.sourceContentStyle = {};
+        this.sourceListLoading = false;
+        this.targetListLoading = false;
+        this.displayNoneStyle = { display: "none" };
```
AxelPeter: No inline style in our JS. Use CSS classes.
Or use `ngShow`, `ngHide` or `ngIf` if necessary :)

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+
+    getTargetContentHeight () {
+        if (this.bulkActionEnabled) {
+            return { height: "calc(100% - 30px)" };
+        }
+        return { height: "100%" };
+    }
+
+    onTargetContentClose () {
+        this.targetCloseIconStyle = this.displayNoneStyle;
```
AxelPeter: We should use `ngClass` in the component template for this, with boolean for example.

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+    getProperty (item) {
+        if (!this.property) {
+            return item;
+        }
+        return this.property.split(".").reduce((prev, curr) => prev ? prev[curr] : undefined, item);
+    }
+
+    isLoading (item) {
+        const uniqueName = this.getProperty(item);
+        if (this.loadingMap[uniqueName]) {
```
AxelPeter: It could be simplified with

```
return !!this.loadingMap[uniqueName];
```
or
```
return angular.isDefined(this.loadingMap[uniqueName]);
```

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+            // remove from target list
+            const newIndex = this.targetList.indexOf(item);
+            this.targetList.splice(newIndex, 1);
+        });
+        if (this.onAdd) {
+            this.onAdd({ items: list })
+                .then(() => {
+                    // all items successfully moved, remove loading
+                    list.forEach(item => this.setLoading(item, false));
+                }, failedItems => {
```
AxelPeter: This could be in the `catch()` ?

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+        return this.$q.when(this.sourceList)
+            .then(items => {
+                this.sourceList = items.data ? items.data : items;
+            })
+            .finally(() => {
+                this.sourceListLoading = false;
+            });
+    }
+
+    loadtargetList () {
```
cbourgois: `loadTargetList`maybe ?

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+            // remove from source list
+            const newIndex = this.sourceList.indexOf(item);
+            this.sourceList.splice(newIndex, 1);
+        });
+        if (this.onAdd) {
+            this.onAdd({ items: list })
+                .then(() => {
+                    // all items successfully moved, remove loading
+                    list.forEach(item => this.setLoading(item, false));
+                }, failedItems => {
```
cbourgois: This could be in the catch() ?

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+            });
+    }
+
+    add (index, item) {
+        if (this.isLoading(item)) {
+            return;
+        }
+        this.sourceList.splice(index, 1);
+        this.targetList.push(item);
+        if (this.onAdd) {
```
cbourgois: Should we have `onBeforeAdd` event instead of undoing everything ?

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+
+    remove (index, item) {
+        if (this.isLoading(item)) {
+            return;
+        }
+        this.targetList.splice(index, 1);
+        this.sourceList.push(item);
+        if (this.onRemove) {
```
cbourgois: Same here (`onBeforeRemove`) ?

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+
+    add (index, item) {
+        if (this.isLoading(item)) {
+            return;
+        }
+        this.sourceList.splice(index, 1);
+        this.targetList.push(item);
+        if (this.onAdd) {
+            this.setLoading(item, true);
+            this.onAdd({ items: [item] })
```
cbourgois: `onAdd`is always called with an object containing a list of items, we could just pass the list ?

packages/oui-dual-list/src/dual-list.filter.js
```diff
@@ -0,0 +1,11 @@
+export default function (items, sourceFilterKey, property) {
+    return items.filter(item => {
+        let value;
+        if (!property) {
+            value = item;
+        } else {
+            value = property.split(".").reduce((prev, curr) => prev ? prev[curr] : undefined, item);
```
cbourgois: Maybe we could use lodash `_.get` ?

AxelPeter: This method can be imported it like this:
```
import { get } from "lodash";
```

packages/oui-dual-list/src/dual-list.html
```diff
@@ -0,0 +1,89 @@
...
+            <div class="oui-dual-list__empty_label" data-ng-if="$ctrl.sourceListLoading">
+                <oui-spinner size="m"></oui-spinner>
+            </div>
+            <div class="oui-dual-list___full_height" data-ng-if="!$ctrl.sourceListLoading && $ctrl.sourceList.length">
+                <div class="oui-dual-list__content_filter">
+                    <input type="text" data-ng-model="$ctrl.sourceFilterKey" ng-model-options="{debounce: 500}"/>
+                    <i class="oui-icon oui-icon-search" aria-hidden="true"></i>
+                </div>
+                <div class="oui-dual-list__content_add_all" data-ng-if="$ctrl.bulkActionEnabled">
+                    <a data-ng-click="$ctrl.addAll()">{{$ctrl.moveAllLabel}}</a>
```
AxelPeter: We should use `class="oui-link"` for links

packages/oui-dual-list/src/dual-list.html
```diff
@@ -0,0 +1,89 @@
...
+        <div class="oui-dual-list__content oui-dual-list___full_height" data-ng-style="$ctrl.targetContentStyle">
+            <div class="oui-dual-list__empty_label" data-ng-if="!$ctrl.targetListLoading && !$ctrl.targetList.length">
+                <p>{{$ctrl.targetListEmptyLabel}}</p>
+            </div>
+            <div class="oui-dual-list__empty_label" data-ng-if="$ctrl.targetListLoading">
+                <oui-spinner size="m"></oui-spinner>
+            </div>
+            <div class="oui-dual-list___full_height" data-ng-if="!$ctrl.targetListLoading && $ctrl.targetList.length">
+                <div class="oui-dual-list__content_remove_all" data-ng-if="$ctrl.bulkActionEnabled">
+                    <a id="dual-list-remove-all"
```
AxelPeter: Here too

packages/oui-dual-list/src/dual-list.html
```diff
@@ -0,0 +1,89 @@
...
+                    <i class="oui-icon oui-icon-search" aria-hidden="true"></i>
+                </div>
+                <div class="oui-dual-list__content_add_all" data-ng-if="$ctrl.bulkActionEnabled">
+                    <a data-ng-click="$ctrl.addAll()">{{$ctrl.moveAllLabel}}</a>
+                </div>
+                <ul class="oui-dual-list__content_list" data-ng-style="$ctrl.sourceListHeight">
+                    <li data-ng-repeat="source in $ctrl.sourceList | ouiDualListFilter:$ctrl.sourceFilterKey:$ctrl.property"
+                        data-ng-click="$ctrl.add($index, source)"
+                        class="oui-dual-list__content_list_item">
+                        <span data-ng-if="!$ctrl.isLoading(source)" class="oui-dual-list__content_list_item_icon">
```
AxelPeter: For accessibility purpose, we should not use `span`.
We can't navigate with the keyboard in the component.

Prefer using `button type="button"` with `ngClick` on it, instead on `li`

packages/oui-dual-list/src/dual-list.html
```diff
@@ -0,0 +1,89 @@
...
+                <div class="oui-dual-list__content_remove_all" data-ng-if="$ctrl.bulkActionEnabled">
+                    <a id="dual-list-remove-all"
+                        name="dual-list-remove-all"
+                        data-ng-click="$ctrl.removeAll()">{{$ctrl.removeAllLabel}}</a>
+                </div>
+                <ul class="oui-dual-list__content_list" data-ng-style="$ctrl.targetListHeight">
+                    <li data-ng-repeat="target in $ctrl.targetList"
+                        data-ng-click="$ctrl.remove($index, target)"
+                        class="oui-dual-list__content_list_item oui-dual-list__content_list_item_target">
+                        <span data-ng-if="!$ctrl.isLoading(target)" class="oui-dual-list__content_list_item_icon">
```
AxelPeter: Same here :)

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+                .catch(() => {
+                    const newIndex = this.sourceList.indexOf(item);
+                    this.sourceList.splice(newIndex, 1);
+                    this.targetList.push(item);
+                })
+                .finally(() => this.setLoading(item, false));
+        }
+    }
+
+    addAll () {
```
AxelPeter: I really think this method (and `removeAll`) can be simplified by merging the source array (filtered or not) to the other, instead of pushing each item at a time in the other. Maybe, is there a reason to do it this way ?

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,247 @@
...
+                this.targetList = items.data ? items.data : items;
+            })
+            .finally(() => {
+                this.targetListLoading = false;
+            });
+    }
+
+    add (index, item) {
+        if (this.isLoading(item)) {
+            return;
```
AxelPeter: This will not pass `eslint`

packages/oui-dual-list/src/dual-list.directive.js
```diff
@@ -0,0 +1,24 @@
+import controller from "./dual-list.controller";
+import template from "./dual-list.html";
+
+export default () => ({
```
AxelPeter: Prefer using angular component instead of a directive, since it will be use with `<oui-dual-list></oui-dual-list>`.
In this case, the file will renamed `dual-list.component.js` :)

packages/oui-dual-list/src/dual-list.controller.js
radireddy: No. items which are loading should not be moved by clicking on add. That's why I loop through all items and pick items which are not in the loading state.

packages/oui-dual-list/src/dual-list.filter.js
radireddy: how to replace "property.split(".").reduce((prev, curr) => prev ? prev[curr] : undefined, item);
" with  _.get ?

packages/oui-dual-list/src/dual-list.controller.js
radireddy: Yes, there is a reason. All items can not be moved blindly when the user clicks add all link. Items which are loading should not be added. That's why I loop through all items and pick items which are not in the loading state.

packages/oui-dual-list/src/dual-list.filter.js
cbourgois: In this case `let value = get(item, property, item);` for replacing all the `if` block

packages/oui-dual-list/src/dual-list.controller.js
cbourgois: I just mean replacing ```this.onAdd({ items: [item] })``` by ```this.onAdd([item])```

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,251 @@
...
+        this.loadingMap = {};
+        this.sourceListLoading = false;
+        this.targetListLoading = false;
+        this.loadSourceList();
+        this.loadTargetList();
+    }
+
+    $postLink () {
+        this.$element.addClass("oui-dual-list");
+        this.sourceCloseIcon = this.$element[0].querySelector(".oui-dual-list__source > .oui-dual-list__header > .oui-dual-list__header_toggle > .oui-dual-list__header_toggle_up");
```
AxelPeter: I think we misunderstood ourself for this one :)

Unless if really necessary (eg. `this.$element.addClass("oui-dual-list");`, to add classname on root element), all style/class management must be done in the HTML template using `ngClass`. We must avoid DOM modifications in a jQuery-like way in our AngularJS code. Generally, we tend to do only DOM manipulations in the `$postLink`, wrapped with `$timeout`.

To show/hide objects, you have directives like `ngIf`, `ngHide` or `ngShow`.
In this way, methods like `onTargetContentClose()`, `onTargetContentOpen()`, `onSourceContentClose()` and `onSourceContentOpen()` should be really simplified.

cf. my other comment :)

packages/oui-dual-list/src/dual-list.controller.js
```diff
@@ -0,0 +1,251 @@
...
+        this.sourceOpenIcon = this.$element[0].querySelector(".oui-dual-list__source > .oui-dual-list__header > .oui-dual-list__header_toggle > .oui-dual-list__header_toggle_down");
+        this.sourceContent = this.$element[0].querySelector(".oui-dual-list__source > .oui-dual-list__content");
+        this.targetCloseIcon = this.$element[0].querySelector(".oui-dual-list__target > .oui-dual-list__header > .oui-dual-list__header_toggle > .oui-dual-list__header_toggle_up");
+        this.targetOpenIcon = this.$element[0].querySelector(".oui-dual-list__target > .oui-dual-list__header > .oui-dual-list__header_toggle > .oui-dual-list__header_toggle_down");
+        this.targetContent = this.$element[0].querySelector(".oui-dual-list__target > .oui-dual-list__content");
+
+        angular.element(this.sourceCloseIcon).addClass("oui-dual-list__display-inline-block");
```
AxelPeter: We really should avoid using classnames with CSS property in its name.
Especially to toggle a display value that could be done in an AngularJS way. 

For example, here for your sourceCloseIcon, we could replace

```
<div class="oui-dual-list__source">
    <div class="oui-dual-list__header">
        [...]
        <span class="oui-dual-list__header_toggle">
            <i class="oui-icon oui-icon-chevron-up oui-dual-list__header_toggle_up"
                data-ng-click="$ctrl.onSourceContentClose()"
                aria-hidden="true"></i>
            <i class="oui-icon oui-icon-chevron-down oui-dual-list__header_toggle_down"
                data-ng-click="$ctrl.onSourceContentOpen()"
                aria-hidden="true"></i>
        </span>
    </div>
    <div class="oui-dual-list__content">
        [...]
    </div>
</div>
```
by
```
<div class="oui-dual-list__source">
    <div class="oui-dual-list__header">
        [...]
        <button class="oui-dual-list__header_toggle"
            ng-click="$ctrl.toggleSource()">
            <span class="oui-icon" ng-class="{
                    'oui-icon-chevron-up': $ctrl.isSourceOpen,
                    'oui-icon-chevron-down': !$ctrl.isSourceOpen
                }">
            </span>
        </button>
    </div>
    <div class="oui-dual-list__content" ng-show="$ctrl.isSourceOpen">
        [...]
    </div>
</div>

toggleSource () {
    this.isSourceOpen = !this.isSourceOpen;
}
```

packages/oui-dual-list/src/dual-list.html
```diff
@@ -0,0 +1,80 @@
...
+            <div class="oui-dual-list__empty-label" data-ng-if="!$ctrl.sourceListLoading && !$ctrl.sourceList.length">
+                <p>{{$ctrl.sourceListEmptyLabel}}</p>
+            </div>
+            <div class="oui-dual-list__empty-label" data-ng-if="$ctrl.sourceListLoading">
+                <oui-spinner size="m"></oui-spinner>
+            </div>
+            <div class="oui-dual-list__content_body" data-ng-if="!$ctrl.sourceListLoading && $ctrl.sourceList.length">
+                <oui-search class="oui-dual-list__content_filter" model="$ctrl.sourceFilterKey" ng-model-options="{debounce: 500}" aria-label="search items">
+                </oui-search>
+                <div class="oui-dual-list__content_add-all" data-ng-if="$ctrl.bulkActionEnabled">
```
AxelPeter: Careful on your BEM naming, for `oui-dual-list__content_add-all`. I don't think `&_add-all` is a modifier.
For example, a more compliant name could be `oui-dual-list__add-all-button`, since it's an element :)

It apply too for all others classnames in this template ;)

packages/oui-dual-list/src/dual-list.controller.js
radireddy: the component has to send an array of items as it supports bulk operations too.

packages/oui-dual-list/src/dual-list.controller.js
radireddy: Good suggestion. I made required changes but I need to add a code in the component to detect device based on width. Hope it is fine.

packages/oui-dual-list/src/dual-list.controller.js
radireddy: Here I need to add a class to the component root element. If I don't set its display to flex. Scroll does not work. Is there any other way? I have wrapped this code in setTimeout.

radireddy:
fixed accessibility issues and review comments

packages/oui-dual-list/src/dual-list.html
```diff
@@ -0,0 +1,80 @@
...
+        </button>
+    </div>
+    <div class="oui-dual-list__content" data-ng-if="$ctrl.isTargetOpen">
+        <div class="oui-dual-list__content_empty" data-ng-if="!$ctrl.targetListLoading && !$ctrl.targetList.length">
+            <p>{{$ctrl.targetListEmptyLabel}}</p>
+        </div>
+        <div class="oui-dual-list__content_empty" data-ng-if="$ctrl.targetListLoading">
+            <oui-spinner size="m"></oui-spinner>
+        </div>
+        <div class="oui-dual-list__content_body" data-ng-if="!$ctrl.targetListLoading && $ctrl.targetList.length">
```
AxelPeter: We're almost done ! There are still issue with the BEM compliance with your CSS classnames in the component.
I see that you often use modifier as an element. 

Here for example with `oui-dual-list__content_body`. `body` is not a modifier for `oui-dual-list__content`.
For example, it could be `oui-dual-list__content-body`, or `oui-dual-list__body`, or `oui-dual-list-content__body` to be compliant.

Please see https://en.bem.info/methodology/naming-convention/ for more informations
In a simple way, `block_modifier` or `block___element_modifier` but not `block___element_element` :)

packages/oui-dual-list/src/dual-list.filter.js
AxelPeter: This could be used for the filter too :)

radireddy: done. used in filter too.

jleveugle:
Hello @radireddy,
Your pull request is now closed due to inactivity.
Feel free to re-open it if you have any update.
Anna Bot for OVH UX <3

jleveugle:
Hello @radireddy,
Your pull request is now closed due to inactivity.
Feel free to re-open it if you have any update.
Anna Bot for OVH UX <3

PR #167, opened 2018-04-19
fix(oui-datagrid): fix reload based on cached rows

JeremyDec:
Fix reload data with cached rows


AxelPeter:
@JeremyDec can you remove the build files for the merge ? Thanks :)

JeremyDec:
@AxelPeter ok it's done

PR #168, opened 2018-04-19
feat(oui-field): add the label popover option

rfougier:


packages/oui-field/README.md
```diff
@@ -45,6 +45,27 @@
 </form>
 ```
 
+### Label Popover
+
+```html:preview
+<form novalidate name="inputForm">
+    <oui-field label="Lastname"
+        help-text="At least 3 chars"
+        label-popover="In the LastName field, enter your family name as the family name appears on your passport. The last name is required to complete the application."
```
neolitec: Later, we'll probably have to propose another solution to get rich content in these popovers.

PR #173, opened 2018-04-25
feat(oui-collapsible): collapsible component addition

ganeshkumar1989:
UU-836
Collapsible component added.

**Linked to**
https://github.com/ovh-ux/ovh-ui-kit/pull/197
https://github.com/ovh-ux/ovh-ui-kit-documentation/pull/94

### DIFF
--- a/packages/oui-collapsible/src/collapsible.controller.js
+++ b/packages/oui-collapsible/src/collapsible.controller.js
@@ -1,7 +1,5 @@
 import { addBooleanParameter } from "@oui-angular/common/component-utils";
 
-const baseClass = "oui-collapsible";
-
 export default class {
     constructor ($attrs, $element, $timeout) {
         "ngInject";
@@ -12,7 +10,6 @@ export default class {
     }
 
     $onInit () {
-        this.baseClass = baseClass;
         addBooleanParameter(this, "expanded");
     }
 
--- a/packages/oui-collapsible/src/collapsible.controller.js
+++ b/packages/oui-collapsible/src/collapsible.controller.js
@@ -1,0 +1,22 @@
+import { addBooleanParameter } from "@oui-angular/common/component-utils";
+
+const baseClass = "oui-collapsible";
+
+export default class {
+    constructor ($attrs, $element, $timeout) {
+        "ngInject";
+
+        this.$attrs = $attrs;
+        this.$element = $element;
+        this.$timeout = $timeout;
+    }
+
+    $onInit () {
+        this.baseClass = baseClass;
+        addBooleanParameter(this, "expanded");
+    }
+
+    toggle () {
+        this.expanded = !this.expanded;
+    }
+}
--- a/packages/oui-collapsible/src/collapsible.controller.js
+++ b/packages/oui-collapsible/src/collapsible.controller.js
@@ -17,6 +17,7 @@ export default class {
         // So we use $timeout to force the $apply
         this.$timeout(() =>
             this.$element.removeAttr("id")
+                .removeAttr("aria-label")
         );
     }
 
--- a/packages/oui-collapsible/src/collapsible.controller.js
+++ b/packages/oui-collapsible/src/collapsible.controller.js
@@ -1,16 +1,23 @@
-import { addBooleanParameter } from "@oui-angular/common/component-utils";
+import { addDefaultParameter } from "@oui-angular/common/component-utils";
 
 export default class {
     constructor ($attrs, $element, $timeout) {
         "ngInject";
-
         this.$attrs = $attrs;
         this.$element = $element;
         this.$timeout = $timeout;
     }
 
     $onInit () {
-        addBooleanParameter(this, "expanded");
+        addDefaultParameter(this, "expanded", false);
+    }
+
+    $postLink () {
+        // Sometimes the digest cycle is done before dom manipulation,
+        // So we use $timeout to force the $apply
+        this.$timeout(() =>
+            this.$element.removeAttr("id")
+        );
     }
 
     toggle () {
--- a/packages/oui-collapsible/src/collapsible.controller.js
+++ b/packages/oui-collapsible/src/collapsible.controller.js
@@ -1,25 +1,37 @@
 import { addDefaultParameter } from "@oui-angular/common/component-utils";
 
 export default class {
-    constructor ($attrs, $element, $timeout) {
+    constructor ($attrs, $element, $scope, $timeout, $window) {
         "ngInject";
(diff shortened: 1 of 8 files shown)


packages/oui-collapsible/src/collapsible.controller.js
```diff
@@ -0,0 +1,19 @@
+import { addBooleanParameter } from "@oui-angular/common/component-utils";
+
+export default class {
+    constructor ($attrs, $element, $timeout) {
+        "ngInject";
+
+        this.$attrs = $attrs;
+        this.$element = $element;
+        this.$timeout = $timeout;
```
AxelPeter: `this.$element` and `this.$timeout` are not used. We can remove them.
But we still need `this.$attrs` for addBooleanParameter`addBooleanParameter.

packages/oui-collapsible/src/collapsible.html
```diff
@@ -0,0 +1,13 @@
+<div class="oui-collapsible" data-ng-class="{'oui-collapsible_open':$ctrl.expanded}">
+    <div class="oui-collapsible__header" data-ng-click="$ctrl.toggle()">
+        <div class="oui-collapsible__header-text" data-ng-bind="::$ctrl.title"></div>
+        <button
```
AxelPeter: This `<button>` seems useless since the click is on the parent `<div>`.
What do you think about replacing `oui-collapsible__header` by a `<button>` ?
```
<button class="oui-collapsible__header" type="button"
    aria-label="{{::$ctrl.ariaLabel}}" aria-expanded="{{$ctrl.expanded}}"
    ng-click="$ctrl.toggle()">
    <span class="oui-collapsible__header-text" ng-bind="::$ctrl.title"></span>
    <span class="oui-icon oui-icon-chevron-down oui-collapsible__toggle-icon" aria-hidden="true"></span>
</button>
```

packages/oui-collapsible/src/collapsible.html
```diff
@@ -0,0 +1,13 @@
+<div class="oui-collapsible" data-ng-class="{'oui-collapsible_open':$ctrl.expanded}">
+    <div class="oui-collapsible__header" data-ng-click="$ctrl.toggle()">
```
AxelPeter: Accessibility issue: we have no clue when the header has the focus (tab navigation). 
We could add an outline like on the navbar for this :)

packages/oui-collapsible/src/collapsible.html
```diff
@@ -0,0 +1,13 @@
+<div class="oui-collapsible" data-ng-class="{'oui-collapsible_open':$ctrl.expanded}">
```
AxelPeter: Unlike in our managers, we don't prefixing with `data-` in our templates

packages/oui-collapsible/src/collapsible.html
```diff
@@ -0,0 +1,13 @@
...
+        <div class="oui-collapsible__header-text" data-ng-bind="::$ctrl.title"></div>
+        <button
+            type="button"
+            aria-label="{{::$ctrl.ariaLabel}}"
+            aria-pressed="{{$ctrl.expanded}}"
+            class="oui-collapsible__toggle-button">
+            <i class="oui-icon oui-icon-chevron-down oui-collapsible__toggle-icon" aria-hidden="true"></i>
+        </button>
+    </div>
+    <div class="oui-collapsible__body" data-ng-transclude></div>
```
AxelPeter: Accessibility issue: we have no `aria-expanded` and `aria-controls`.
Please see [here](https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions#Example_1:_Using_a_button_to_collapse_and_expand_a_region) for more informations :)

packages/oui-collapsible/src/collapsible.html
```diff
@@ -0,0 +1,13 @@
+<div class="oui-collapsible" data-ng-class="{'oui-collapsible_open':$ctrl.expanded}">
+    <div class="oui-collapsible__header" data-ng-click="$ctrl.toggle()">
+        <div class="oui-collapsible__header-text" data-ng-bind="::$ctrl.title"></div>
+        <button
+            type="button"
+            aria-label="{{::$ctrl.ariaLabel}}"
+            aria-pressed="{{$ctrl.expanded}}"
```
AxelPeter: I think this should be `aria-expanded` instead

packages/oui-collapsible/src/collapsible.controller.js
ganeshkumar1989: Now the component accepts an id, which is required for aria-controls attribute. This id must be removed from the original directive and hence these dependencies are now required :)

packages/oui-collapsible/src/collapsible.html
ganeshkumar1989: This is done :)

packages/oui-collapsible/src/collapsible.html
ganeshkumar1989: The data- prefix has now been removed :)

packages/oui-collapsible/src/collapsible.html
ganeshkumar1989: That is a very good idea :)
Expanding a panel on button click sounds good. But is having a button as a header a good idea? Please let me know your thoughts on this.
Another possibility is to remove the button and have the icon in oui-collapsible__header

AxelPeter: For me, it's not a problem, as long as we don't have a click on a unfocusable tag. 
We could still have a wrapper around the button if necessary, but you'll need the `aria-controls` attribute for a11y:
eg.
```
<div class="oui-collapsible__header">
    <button class="oui-collapsible__header-button" type="button"
        aria-label="{{::$ctrl.ariaLabel}}" 
        aria-expanded="{{$ctrl.expanded}}"
        aria-controls="collapsibleContent1"
        ng-click="$ctrl.toggle()">
        <span class="oui-collapsible__header-text" ng-bind="::$ctrl.title"></span>
        <span class="oui-icon oui-icon-chevron-down oui-collapsible__toggle-icon" aria-hidden="true"></span>
    </button>
</div>
<div class="oui-collapsible__body" id="collapsibleContent1">
    [...]
</div>
```

ganeshkumar1989: Understood. After some thought, it seems like a button for the header is a better idea. I have done this change :)
Thanks again for this idea

PR #174, opened 2018-04-26
Feature/oui clipboard

varun257:
UU-832

feature: oui clipboard component

change oui tooltip from one time binding to one way binding so as to reuse tooltip
add new component clipboard

Demo: http://feature_oui_clipboard_build.ui-kit.ovh/#!/oui-angular/clipboard

### DIFF
--- a/packages/oui-clipboard/src/clipboard.component.js
+++ b/packages/oui-clipboard/src/clipboard.component.js
@@ -4,7 +4,7 @@ import template from "./clipboard.html";
 export default {
     template,
     controller,
-    controllerAs: "$clipctrl",
+    controllerAs: "$ctrl",
     bindings: {
         text: "@?"
     }
--- a/packages/oui-clipboard/src/clipboard.component.js
+++ b/packages/oui-clipboard/src/clipboard.component.js
@@ -6,8 +6,6 @@ export default {
     controller,
     controllerAs: "$clipctrl",
     bindings: {
-        label: "<?",
-        id: "@?",
         text: "@?"
     }
 };
--- a/packages/oui-clipboard/src/clipboard.component.js
+++ b/packages/oui-clipboard/src/clipboard.component.js
@@ -1,0 +1,13 @@
+import controller from "./clipboard.controller";
+import template from "./clipboard.html";
+
+export default {
+    template,
+    controller,
+    controllerAs: "$clipctrl",
+    bindings: {
+        label: "<?",
+        id: "@?",
+        text: "@?"
+    }
+};
--- a/packages/oui-clipboard/src/clipboard.component.js
+++ b/packages/oui-clipboard/src/clipboard.component.js
@@ -6,8 +6,8 @@ export default {
     controller,
     controllerAs: "$ctrl",
     bindings: {
-        text: "@?",
         name: "@?",
-        id: "@?"
+        id: "@?",
+        model: "<?"
     }
 };
--- a/packages/oui-clipboard/src/clipboard.component.js
+++ b/packages/oui-clipboard/src/clipboard.component.js
@@ -4,10 +4,9 @@ import template from "./clipboard.html";
 export default {
     template,
     controller,
-    controllerAs: "$ctrl",
     bindings: {
         name: "@?",
         id: "@?",
-        model: "<?"
+        model: "="
     }
 };
--- a/packages/oui-clipboard/src/clipboard.component.js
+++ b/packages/oui-clipboard/src/clipboard.component.js
@@ -6,6 +6,8 @@ export default {
     controller,
     controllerAs: "$ctrl",
     bindings: {
-        text: "@?"
+        text: "@?",
+        name: "@?",
+        id: "@?"
     }
 };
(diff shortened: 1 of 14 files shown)


packages/oui-clipboard/src/clipboard.html
```diff
@@ -0,0 +1,12 @@
+<div class="oui-clipboard">
+    <div class="oui-clipboard__container">
+        <input class="oui-clipboard__input text-truncate"
+            type="text"
+            data-ng-focus="$clipctrl.onTextFocus($event)"
+            data-ng-value="$clipctrl.text"
+            data-oui-tooltip="{{$clipctrl.tooltipText}}"
+            readonly>
+        <span class="oui-clipboard__icon" data-ng-if="$clipctrl.status === 'initial'"><i class="oui-icon oui-icon-copy-normal" aria-hidden="true"></i></span>
```
AxelPeter: Since the icon must be clickable, this should be in a `<button>`

packages/oui-clipboard/src/clipboard.html
```diff
@@ -0,0 +1,12 @@
+<div class="oui-clipboard">
```
AxelPeter: This class could be added on the root component `<oui-clipboard>`, to avoid this DIV.
See our use of `$postLink` for that :)

packages/oui-clipboard/src/clipboard.html
```diff
@@ -0,0 +1,12 @@
+<div class="oui-clipboard">
+    <div class="oui-clipboard__container">
```
AxelPeter: Why not use `oui-input-group oui-input-group_button` ? We could even go with a new modifier if necessary (eg. `oui-input-group_clipboard`)

packages/oui-clipboard/src/clipboard.provider.js
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Set the translations
+     * @param {Object} translations a map of translations
+     */
+    setTranslations (translations) {
+        this.translations = translations;
+        return this;
+    }
+
+    setStatus (status) {
```
AxelPeter: Is this really necessary ? Do we have already a use of this method ?

packages/oui-clipboard/src/clipboard.provider.js
```diff
@@ -0,0 +1,42 @@
...
+        this.status = status;
+        return this;
+    }
+
+    setAction (copy) {
```
AxelPeter: Same for this one too :)

packages/oui-clipboard/src/clipboard.component.js
```diff
@@ -0,0 +1,11 @@
+import controller from "./clipboard.controller";
+import template from "./clipboard.html";
+
+export default {
+    template,
+    controller,
+    controllerAs: "$clipctrl",
```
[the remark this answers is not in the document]
varun257: The tooltip has $ctrl as controller because of which i lose scope of the parent controller. So had to rename this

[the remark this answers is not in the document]
AxelPeter: Ok. In this case, I think we should rename the controller of the tooltip instead with `$tooltipCtrl`.

packages/oui-clipboard/src/clipboard.provider.js
varun257: using these functions in spec for test cases

varun257:
Hi @AxelPeter, Have done the changes as suggested. Please check if this is okay now

packages/oui-clipboard/src/clipboard.html
```diff
@@ -1,10 +1,13 @@
...
     <input class="oui-input"
         type="text"
         ng-attr-id="{{::$ctrl.id}}"
         ng-attr-name="{{::$ctrl.name}}"
-        ng-click="$ctrl.onClick()"
         ng-model="::$ctrl.model"
         contenteditable>
-    <button class="oui-button"><i class="oui-icon oui-icon-copy-normal" aria-hidden="true"></i></button>
+    <button class="oui-button">
+        <span class="oui-icon oui-icon-copy-normal"></span>
```
FredericEspiau: `aria-hidden="true"`

jleveugle: We should have a description on this button, like "Copy" or something like this :)

FredericEspiau: Yep I already told him face to face

packages/oui-clipboard/src/clipboard.provider.js
```diff
@@ -4,7 +4,7 @@ export default class {
         this.translations = {
             copyToClipboardLabel: "Copy to Clipboard",
             copiedLabel: "Copied",
-            notSupported: "Not supported. Please copy the text manually"
+            notSupported: "Copy to Clipboard not supported. Please copy the text manually"
```
FredericEspiau: `clipboard`

AxelPeter: Was like this in the guidelines :)

packages/oui-clipboard/src/clipboard.html
AxelPeter: I've added an `aria-label` with the tooltip text "Copy to clipboard"

PR #175, opened 2018-04-26
feat(oui-modal): add type attribute to handle icon modal

whispyy:
add type attribute that allows to handle icon : info, warning, success ..etc

### DIFF
--- a/packages/oui-modal/src/modal.html
+++ b/packages/oui-modal/src/modal.html
@@ -7,7 +7,12 @@
       ng-click="$ctrl.onDismiss()">
     </button>
   </div>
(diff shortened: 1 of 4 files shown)


packages/oui-modal/src/modal.html
```diff
@@ -7,7 +7,12 @@
       ng-click="$ctrl.onDismiss()">
     </button>
   </div>
-  <div class="oui-modal__body oui-modal__body_no-icon">
+  <div class="oui-modal__body"
+       ng-class="::{'oui-modal__body_no-icon': !$ctrl.type}">
+    <i  ng-if="::!!$ctrl.type"
```
rfougier: Prefer `<span>`

PR #176, opened 2018-04-27
fix(oui-select-picker): fix matched value display

rfougier:


packages/oui-select-picker/src/select-picker.html
```diff
@@ -42,7 +42,7 @@
                         placeholder="{{:: $ctrl.placeholder || ' - ' }}"
                         data-align="end"
                         ng-if="$ctrl.values.length > 1">
-                <span ng-bind="$item.name"></span>
+                <span ng-bind=":: $ctrl.$scope.$parent.getItemValue($item, $ctrl.match)"></span>
```
neolitec: Okay because it's internal code :)

PR #177, opened 2018-04-30
onClick is optional on action-menu-item

cbourgois:
`onClick` should be optional on `action-menu-item`

### DIFF
--- a/packages/oui-action-menu/src/action-menu-item.component.js
+++ b/packages/oui-action-menu/src/action-menu-item.component.js
@@ -8,6 +8,6 @@ export default {
         ariaLabel: "@?",
         href: "@?",
         external: "<?",
-        onClick: "&"
+        onClick: "&?"
     }
 };


neolitec:
Does it create issue?

JeremyDec:
Hello @neolitec, the issue is that it's impossible for an action item, that is a href link (and not a button), to be disabled. (missing on-click mandatory option prevent disabling the link dynamically)

neolitec:
@JeremyDec a link can't be disabled and we are

neolitec:
Ok, I got it. You want to disable a link. But you can't easily do that and getting onClick optional does not change anything in that.
But maybe you could use `onClick` on the `a` element to `preventDefault` action if `disabled` is `true`.

AxelPeter:
No need of this modification. We use `"&"` binding because it use `angular.noop()` if undefined. 
With these change, we'll need a check before calling the function.

PR #178, opened 2018-05-02
feat(oui-tooltip): add possibility to display html

JeremyDec:


packages/oui-tooltip/src/tooltip.directive.js
```diff
@@ -4,7 +4,8 @@ export default () => ({
     restrict: "A",
     bindToController: {
         text: "@ouiTooltip",
-        placement: "@?ouiTooltipPlacement" // values: top|top-start|top-end|bottom|bottom-start|bottom-end (default: top)
+        placement: "@?ouiTooltipPlacement", // values: top|top-start|top-end|bottom|bottom-start|bottom-end (default: top)
+        isHtml: "<?ouiTooltipIsHtml"
```
neolitec: It could be html all the way. But keep in mind that tooltip should not contain complex markup. Only simple text modifiers like bold, italic, colors, etc...

neolitec: If you want to show a list (ui/li) in a floating box over the UI, you should use a Popover (http://master.ui-kit.ovh/#!/oui-angular/popover).

JeremyDec: Ok, I used popover component, so I close this PR

PR #182, opened 2018-05-16
fix(oui-modal): allow 2 way binding on some attributes

Jisay:
As I want to load some informations during modal controller initialization, it can happen that I need to display an error when the action in the modal is not possible.
In this case button could also be different, so I need one time binding to be removed on `type`, `primaryLabel` and `secondaryLabel` attributes.


packages/oui-modal/src/modal.component.js
```diff
@@ -7,12 +7,12 @@ export default {
     bindings: {
         title: "@?",
         primaryAction: "&?",
-        primaryLabel: "@?",
+        primaryLabel: "<?",
```
AxelPeter: No need to change the binding, it will break the existing oui-modal on managers

PR #185, opened 2018-05-18
feat: oui skeleton

varun257:
add oui skeleton component
add examples and test case

cui skeleton to oui

JIRA: UU-845

### DIFF
--- a/packages/oui-skeleton/README.md
+++ b/packages/oui-skeleton/README.md
@@ -19,8 +19,18 @@
 <oui-skeleton size="auto"></oui-skeleton>
 ```
 
+### Randomized width
+```html:preview
+<oui-skeleton randomized></oui-skeleton>
+<oui-skeleton randomized></oui-skeleton>
+<oui-skeleton randomized></oui-skeleton>
+<oui-skeleton randomized></oui-skeleton>
+<oui-skeleton randomized></oui-skeleton>
+```
+
 ## API
 
-| Attribute     | Type                    | Binding | One-time Binding | Values                     | Default               | Description
-| ----          | ----                    | ----    | ----             | ----                       | ----                  | ----
-| size          | String                  | @?      | yes              | `xs`, `s`, `m`, `l`, `xl`  | `auto`                | Skeleton size
+| Attribute     | Type      | Binding | One-time Binding | Values                     | Default   | Description
+| ----          | ----      | ----    | ----             | ----                       | ----      | ----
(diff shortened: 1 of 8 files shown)


packages/oui-skeleton/src/skeleton.component.js
```diff
@@ -0,0 +1,13 @@
+import controller from "./skeleton.controller";
+import template from "./skeleton.html";
+
+export default {
+    controller,
+    controllerAs: "$ctrl",
```
radireddy: this is the default value, you can remove it

packages/oui-skeleton/src/skeleton.html
```diff
@@ -0,0 +1,6 @@
+<div class="oui-skeleton">
+    <span class="oui-skeleton__loader"
+      ng-show="$ctrl.loading"
+      data-ng-style="{width: $ctrl.width}"></span>
+    <div ng-hide="$ctrl.loading" ng-transclude></div>
```
radireddy: use ng-if

packages/oui-skeleton/src/skeleton.controller.js
```diff
@@ -0,0 +1,15 @@
+import { addDefaultParameter } from "@oui-angular/common/component-utils";
+
+export default class {
+    constructor ($attrs) {
+        "ngInject";
+
+        this.$attrs = $attrs;
+        this.max = 100;
```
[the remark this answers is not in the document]
varun257: Yup this should be good. Better than choosing random width. Will do this

packages/oui-skeleton/README.md
```diff
@@ -0,0 +1,26 @@
...
+<component-status cx-design="complete" ux="rc"></component-status>
+
+## Usage
+
+### Default
+```html:preview
+<oui-skeleton></oui-skeleton>
+```
+
+### Size
```
neolitec: Maybe size is not the thing we want.
Typically the skeleton graphically represents a text we don't know in advance. So the size should be... randomized :)
Maybe the right parameters here should be "min-width" and eventually "width" / "auto" if we want to force the width... ?
@AxelPeter what do you think about?

AxelPeter: Nice idea. But I don't think it's the case for all uses.
We can even have both with an attribute "randomized" for example :)

In fact, I'm not a fan of replacing `style="width: 80%"` just by `width="80%"`. That's why by default, I let the skeleton with `auto` to let the user set its width.

I'll check to add your idea

PR #187, opened 2018-05-22
feat(oui-status): status component implementation

ganeshkumar1989:
UU-844
Added oui-status component

**Linked To:**
https://github.com/ovh-ux/ovh-ui-kit-documentation/pull/109


AxelPeter:
Sorry but there is no need of a component for this one. If the component puropose is for replacing a `span` with a class by a `oui-status` with an attribute, the style should be sufficient.
Thanks for your contribution.

PR #188, opened 2018-05-23
fix(oui-datagrid): wait row promise before displaying cell template

frenautvh:
Wait for row promise to be resolved before displaying row template.

For example :
```
<oui-column>
    <span data-ng-if="$row.foo">Foo</span>
    <span data-ng-if="!$row.foo">Bar</span>
</oui-column>
```

Will display "Bar" during on the fly row loading which is not intended (we expect cell to be left blank). This fix wait the promise to be resolved and then compile the cell template.

### DIFF
--- a/packages/oui-datagrid/src/cell/cell.controller.js
+++ b/packages/oui-datagrid/src/cell/cell.controller.js
@@ -12,7 +12,13 @@ export default class {
 
         this.$element.css("display", "block");
 
-        this._compileCell();
+        if (this.row && this.row.$promise) {
+            this.row.$promise.finally(() => {
+                this._compileCell();
+            });
+        } else {
+            this._compileCell();
+        }
     }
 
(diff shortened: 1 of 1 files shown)


packages/oui-datagrid/src/cell/cell.controller.js
```diff
@@ -12,7 +12,13 @@ export default class {
 
         this.$element.css("display", "block");
 
-        this._compileCell();
+        if (this.row && this.row.$promise) {
```
FredericEspiau: Suggestion:

```js
this.$q
	.when(null)
	.then(() => _(this.row).get("$promise"))
	.finally(() => this._compileCell());
```

frenautvh: i like your suggestion but i think an "if" is more straightforward and easier to read in that case

PR #189, opened 2018-05-23
fix(oui-datagrid): add refresh & update events

frenautvh:
Add two events : "datagrid-refresh" & "datagrid-update-rows". In order to be able to refresh and alter rows while using on the fly row loading.

For example :
```js
// From another controller
$scope.$broadcast("datagrid-update-rows", (rows) => {
    const user = _.find(rows, { name: "foo" });
    if (user) {
        user.name = "bar"; // alter row
    }
});
```

### DIFF
--- a/packages/oui-datagrid/src/datagrid.controller.js
+++ b/packages/oui-datagrid/src/datagrid.controller.js
@@ -80,16 +80,45 @@ export default class DatagridController {
         const builtColumns = this.buildColumns();
         this.previousRows = this.columns;
 
-        if (this.rowsLoader) {
-            this.paging = this.ouiDatagridPaging.createRemote(this.columns, builtColumns.currentSorting, this.pageSize, this.rowLoader, this.rowsLoader);
-            this.refreshData(() => this.paging.setOffset(1));
-        } else {
-            this.paging = this.ouiDatagridPaging.createLocal(this.columns, builtColumns.currentSorting, this.pageSize, this.rowLoader, this.rows);
+        const refreshDatagrid = () => {
+            if (this.rowsLoader) {
+                this.paging = this.ouiDatagridPaging.createRemote(this.columns, builtColumns.currentSorting, this.pageSize, this.rowLoader, this.rowsLoader);
+                this.refreshData(() => this.paging.setOffset(1));
+            } else {
+                this.paging = this.ouiDatagridPaging.createLocal(this.columns, builtColumns.currentSorting, this.pageSize, this.rowLoader, this.rows);
 
-            if (this.rows) {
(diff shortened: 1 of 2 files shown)


packages/oui-datagrid/src/datagrid.controller.js
```diff
@@ -80,16 +80,45 @@ export default class DatagridController {
...
+                    updater(this.displayedRows);
+                    const rowDiff = this.displayedRows.length - rowCount;
+
+                    // In case rows have been removed/added, updates paging total count
+                    if (this.paging) {
+                        this.paging.totalCount += rowDiff;
+                    }
+                }
+            };
+            if (this.refreshDatagridPromise) {
```
FredericEspiau: You could check if `this. refreshDatagridPromise ` is a Promise

packages/oui-datagrid/src/datagrid.controller.js
```diff
@@ -80,16 +80,45 @@ export default class DatagridController {
...
+        this.$scope.$on("datagrid-refresh", () => {
+            if (!this.refreshDatagridPromise) {
+                refreshDatagrid();
+            }
+        });
+        refreshDatagrid();
+
+        this.$scope.$on("datagrid-update-rows", (ev, updater) => {
+            const doUpdate = () => {
+                if (angular.isFunction(updater)) {
```
FredericEspiau: Don't you have `_` instead ? We should stop switching between `angular` and `_`

packages/oui-datagrid/src/datagrid.controller.js
```diff
@@ -80,16 +80,45 @@ export default class DatagridController {
...
+                    const rowCount = this.displayedRows.length;
```
FredericEspiau: `numberOfRowsBeforeUpdate` ? 😄

packages/oui-datagrid/src/datagrid.controller.js
antleblanc: use `angular` built-in methods otherwise lodash.

frenautvh:
Discussing with @neolitec & @jimmyfortinx : a DatagridService will be created in order to refresh the grid or update rows.
Some drawback of using events : broadcasting will trigger a refresh for every datagrid in the same scope. Managing datagrid by id with a service has the advantage of targeting a specific datagrid.

packages/oui-datagrid/src/datagrid.controller.js
FredericEspiau: Why ? It's the first time I've heard about that, is this a convention ?

PR #193, opened 2018-05-25
chore(oui-calendar): bump version of flatpickr to v4.5.0

AxelPeter:
https://github.com/flatpickr/flatpickr/releases/tag/v4.5.0
Resolves the logs `Error{}` when launching the unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
   },
   "dependencies": {
     "escape-string-regexp": "^1.0.5",
-    "popper.js": "^1.12.9"
+    "popper.js": "^1.12.9",
+    "flatpickr": "4.5.0"
   },
   "devDependencies": {
     "angular": "~1.6.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "eslint-loader": "^1.9.0",
     "eventsource-polyfill": "^0.9.6",
     "express": "^4.15.3",
-    "flatpickr": "4.4.4",
```
antleblanc: Existing any reason of have `flatpickr` as a `devDependencies`? I took `popper.js` as an example which is as a `dependencies` instead.

AxelPeter: Yes, it was because of this import: https://github.com/ovh-ux/ovh-ui-angular/blob/master/packages/oui-calendar/src/calendar.controller.js

But it's the same for Popper.js. I'll add it as a dependency.

PR #209, opened 2018-06-19
feat(oui-header-tabs): add header-tabs component

rfougier:
`/#!/ovh-ui-angular/header-tabs` @ `ovh_ui_kit_documentation-feat_oui_header_tabs_staging`

**Linked to**
https://github.com/ovh-ux/ovh-ui-kit/pull/221
https://github.com/ovh-ux/ovh-ui-kit-documentation/pull/121


packages/oui-header-tabs/README.md
```diff
@@ -0,0 +1,68 @@
+# Header tabs
+
+<component-status cx-design="complete" ux="rc"></component-status>
+
+## Usage
+
+### Normal
+
+```html:preview
+<div class="oui-doc-preview-only-keep-children">
```
AxelPeter: What's the purpose of this `<div>` ?
It seems useless, since it has no style or init.

packages/oui-header-tabs/src/header-tabs-divider.directive.js
```diff
@@ -0,0 +1,11 @@
+import template from "./header-tabs-divider.html";
+
+export default () => {
+    "ngInject";
+
+    return {
+        restrict: "E",
+        template,
+        replace: true
```
AxelPeter: Sadly, it's deprecated. A previous decision was made to not use it. 
We should use another way like `.replaceWith()` for example: https://docs.angularjs.org/api/ng/function/angular.element

packages/oui-header-tabs/src/header-tabs-dropdown.directive.js
```diff
@@ -0,0 +1,15 @@
+import template from "./header-tabs-dropdown.html";
+
+export default () => {
+    "ngInject";
+
+    return {
+        restrict: "E",
+        template,
+        replace: true,
```
AxelPeter: Same here

PR #242, opened 2018-07-31
feat(oui-datagrid): add selectable rows

frenautvh:
UK-42

### DIFF
--- a/packages/oui-datagrid/README.md
+++ b/packages/oui-datagrid/README.md
@@ -91,10 +91,10 @@
 Clicked row action 1: <span ng-if="$ctrl.action1Row">{{$ctrl.action1Row.lastName}}, {{$ctrl.action1Row.firstName}}</span>
 ```
 
-### Global Actions
+### Selectable rows
 
 ```html:preview
-<oui-datagrid rows="$ctrl.data" page-size="5" global-actions>
+<oui-datagrid rows="$ctrl.data" page-size="5" selectable-rows>
   <oui-column title="'First name'" property="firstName" sortable="asc"></oui-column>
   <oui-column title="'Last name'" property="lastName" sortable></oui-column>
   <oui-column title="'Mother'" property="parents.mother.lastName" sortable>
@@ -688,6 +688,7 @@ call `rows-loader` and then a `row-loader` call for each line.
 | `rows-loader`                  | function        | &?      | yes              |                           |                     | gets all rows (returns a promise with all rows)                    |
 | `row-loader`                   | function        | &?      | yes              |                           |                     | gets row details (returns a promise with details)                  |
 | `customizable`                 | boolean         | <?      |                  |                           | false               | if the datagrid is customizable                                    |
+| `selectable-rows`              | boolean         | <?      |                  |                           | false               | if rows can be selected
 | `columns-parameters`           | array           | <?      |                  |                           | undefined           | columns parameters (see below)                                     |
 | `on-columns-parameters-change` | function        | &       |                  |                           |                     | triggered on column parameter change when datagrid is customizable |
 
--- a/packages/oui-datagrid/README.md
+++ b/packages/oui-datagrid/README.md
@@ -91,6 +91,38 @@
 Clicked row action 1: <span ng-if="$ctrl.action1Row">{{$ctrl.action1Row.lastName}}, {{$ctrl.action1Row.firstName}}</span>
 ```
 
+### Global Actions
+
+```html:preview
+<oui-datagrid rows="$ctrl.data" page-size="5" global-actions>
+  <oui-column title="'First name'" property="firstName" sortable="asc"></oui-column>
+  <oui-column title="'Last name'" property="lastName" sortable></oui-column>
+  <oui-column title="'Mother'" property="parents.mother.lastName" sortable>
+    {{$row.parents.mother.lastName}}, {{$row.parents.mother.firstName}}
+  </oui-column>
+  <oui-column title="'Father'" property="parents.father.lastName" sortable>
+    {{$row.parents.father.lastName}}, {{$row.parents.father.firstName}}
+  </oui-column>
+  <oui-column title="'Email'" property="email" sortable>
+    <a href="mailto:{{$value}}">{{$ctrl.label}}: {{$value}}</a>
+  </oui-column>
+  <oui-column title="'Phone'" property="phone"></oui-column>
+  <oui-column title="'Birth'" property="birth" sortable>
+    {{$value|date:short}}
+  </oui-column>
+  <oui-column title="'Selected'">
+    <span>{{ $selected }}</span>
+  </oui-column>
+  <oui-action-menu align="end" compact>
+      <oui-action-menu-item text="Some action" disabled="$selected" on-click="">
+      </oui-action-menu-item>
+  </oui-action-menu>
+  <extra-top>
+    <pre>You have selected {{ $selectedRows.length }} row(s).</pre>
+  </extra-top>
+</oui-datagrid>
+```
+
 ### Empty datagrid
 
 ```html:preview
--- a/packages/oui-datagrid/README.md
+++ b/packages/oui-datagrid/README.md
@@ -111,10 +111,10 @@ Clicked row action 1: <span ng-if="$ctrl.action1Row">{{$ctrl.action1Row.lastName
     {{$value|date:short}}
   </oui-column>
   <oui-column title="'Selected'">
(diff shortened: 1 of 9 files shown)


packages/oui-datagrid/README.md
```diff
@@ -91,6 +91,38 @@
 Clicked row action 1: <span ng-if="$ctrl.action1Row">{{$ctrl.action1Row.lastName}}, {{$ctrl.action1Row.firstName}}</span>
 ```
 
+### Global Actions
+
+```html:preview
+<oui-datagrid rows="$ctrl.data" page-size="5" global-actions>
```
FredericEspiau: Wouldn't it be best to name the parameter "selectable-rows" ?

packages/oui-datagrid/src/datagrid.controller.js
```diff
@@ -271,6 +273,9 @@ export default class DatagridController {
             this.displayedRows = DatagridController.createEmptyRows(this.paging.getCurrentPageSize());
         }
 
+        this.selectedRows = this.selectedRows.map(() => false);
```
FredericEspiau: Can't this contain the actual rows instead of an array of booleans ? Or something to identify the rows ?

```
{
	id: rowId,
	isSelected: false
}
```
If we keep the array of booleans system, then it should be names something like `rowSelectionStatus`

packages/oui-datagrid/src/datagrid.controller.js
```diff
@@ -313,6 +318,37 @@ export default class DatagridController {
...
+    getSelectedRows () {
+        return this.selectedRows.reduce((result, selected, index) => {
+            if (selected) {
+                result.push(this.displayedRows[index]);
+            }
+            return result;
+        }, []);
+    }
+
+    onSelectRow (index, selected) {
```
FredericEspiau: `onRowSelection` ? Still feels wrong as this method is called even when the row is unselected

FredericEspiau: `selected` could be named `isSelected`, so we know it's a boolean

packages/oui-datagrid/src/datagrid.controller.js
```diff
@@ -313,6 +318,37 @@ export default class DatagridController {
...
+        const rowCount = this.displayedRows.length;
+        this.selectedRows[index] = selected;
+        const selectedRowsCount = this.getSelectedRows().length;
+
+        if (selectedRowsCount === rowCount) {
+            this.selectAllRows = true;
+        } else if (selectedRowsCount === 0) {
+            this.selectAllRows = false;
+        } else {
+            this.selectAllRows = null;
```
FredericEspiau: Is this the global comportement for tri-select box ? `null` is `some items are selected, some are not`? Because, if it's not already the selected behavior, this would require a something else than a boolean then

packages/oui-datagrid/src/datagrid.controller.js
```diff
@@ -313,6 +318,37 @@ export default class DatagridController {
...
+        }
+    }
+
+    toggleSelectAllRows (modelValue) {
```
FredericEspiau: `currentSelectionStatus` ?

FredericEspiau:
@frenautvh I've only commented because I feel most point would require a discussion

packages/oui-datagrid/src/datagrid.controller.js
frenautvh: onToggleSelectRow would be better i think

packages/oui-datagrid/src/datagrid.controller.js
frenautvh: yep checkbox component currently use "null"

packages/oui-datagrid/README.md
frenautvh: sounds good

packages/oui-datagrid/src/datagrid.controller.js
frenautvh: since every checkbox needs to have an ng-model i still have to created N booleans for binding ; that's why it made sense to use an array of booleans (i reuse the array)

packages/oui-datagrid/README.md
AxelPeter: Don't forget to add the details in the API table below for this attribute

packages/oui-datagrid/src/cell/cell.controller.js
```diff
@@ -19,6 +19,11 @@ export default class {
         } else {
             this._compileCell();
         }
+
+        this.index = this.index || 0;
```
AxelPeter: Should be in `$onInit`

packages/oui-datagrid/README.md
FredericEspiau: Were it only me, I would have named it `allows-row-selection` but I know you guys won't like it 😄

packages/oui-datagrid/src/datagrid.controller.js
AxelPeter: In this case, `toggleRowSelection` seems more appropriate

packages/oui-datagrid/src/datagrid.controller.js
AxelPeter: We could keep the same logic as above, `toggleRowsSelection`

packages/oui-datagrid/README.md
AxelPeter: In this way, we keep our logic of *able attributes

ovh-cds:
CDS Report yarn-build-ui-kit#1152.0 ✘  
* build
  * build ✘

ovh-cds:
CDS Report yarn-build#987.0 ✘  
* build &amp; test
  * build ✘  
  * test ✘

packages/oui-datagrid/src/datagrid.controller.js
euhmeuh: We could react to the checkbox's `on-change` event and add the corresponding row to the array of rows.

frenautvh: i think your suggestion would make the code "faster" : no reduce in getSelectedRows since you access array directly ; downside is that you need an additionnal array to store rows which is less clean (you still have to give an ngModel to checkbox so you also need an array of booleans).
did i understand you correcly ?

packages/oui-datagrid/src/cell/cell.controller.js
```diff
@@ -19,6 +23,10 @@ export default class {
         } else {
             this._compileCell();
         }
+
+        this.cellScope.$watch(() => this.datagridCtrl.selectedRows[this.index], (isSelected) => {
+            this.cellScope.$isSelected = isSelected || false;
```
JeremyDec: `this.cellScope.$isSelected = !!isSelected` is possible no ?

frenautvh: `!!` is harder to read by someone not used to this js "trick" but yes it would fit the job

frenautvh:
@antleblanc @euhmeuh @FredericEspiau 
Thanks for your comments, i know that there is more than one way to do one thing. What i don't know is why you choose your way over mine and that's the interesting part so please tell me why when you're suggesting alternatives! :) Otherwise we could comment every line of code and never make a change because everyone is right

AxelPeter:
@frenautvh  The right way is the way that works and that keeps the same logic in the project (naming, structure, etc.). That's why I'm here to take the lead on those decisions.
You could still add comments on your choices to explain why you do it that way, and why it's the better choice in your opinion ;)

frenautvh:
@AxelPeter  I'm targeting that kind of comments :
 - "Can't this contain the actual rows instead of an array of booleans ? Or something to identify the rows ?"
 - "this.cellScope.$isSelected = !!isSelected is possible no ?"
- "We could react to the checkbox's on-change event and add the corresponding row to the array of rows."
I know we could do that but WHY would we? I think the reviewer should at least gives an hint otherwise it's just noise in the PR review.

Example :
 - "We could react to the checkbox's on-change event because in order to avoids to generate array of row each time the binding is updated"
 - "You could do this.cellScope.$isSelected = !!isSelected; i think the syntax is more readable"

AxelPeter:
I know, that's why I suggests some methods for working together (it works with our previous team) ;)

packages/oui-datagrid/src/cell/cell.controller.js
JeremyDec: Ok :)

PR #253, opened 2018-08-23
feat(oui-calendar): add enableTime option

Jisay:
Add to the component the enableTime option to display a time picker if needed.
See enableTime option of flatpickr for more information: https://flatpickr.js.org/options/


AxelPeter:
Since it's a feature, could you target the `develop` branch ?

PR #259, opened 2018-09-07
feat(oui-message): add dismissable attribute

AxelPeter:
http://angular_feature_closable_message.ui-kit.ovh/#!/oui-angular/message

* Add an attribute to force the dismissable state

### DIFF
--- a/packages/oui-message/src/message.controller.js
+++ b/packages/oui-message/src/message.controller.js
@@ -1,4 +1,4 @@
-import { addBooleanParameter } from "@oui-angular/common/component-utils";
+import { addBooleanParameter, addDefaultParameter } from "@oui-angular/common/component-utils";
 
 export default class {
     constructor ($attrs) {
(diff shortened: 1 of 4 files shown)


packages/oui-message/src/message.controller.js
```diff
@@ -8,14 +8,13 @@ export default class {
     }
 
     $onInit () {
+        // Guidelines default value for dismissable attribute
+        addDefaultParameter(this, "dismissable", this.type === "info" || this.type === "success");
```
FredericEspiau: Could have an array of types that are dissimisable by default

AxelPeter: What's your proposition ?

FredericEspiau: `message.constants.js`

```js
angular
	.module("moduleName")
	.constant("MESSAGE_CONSTANTS", () => {
		TYPES: {
			info: {
				dismissableByDefault: true
			},
			...
		}
})
```

`message.controller.js`

```js
addDefaultParameter(this, "dismissable", this.MESSAGE_CONSTANTS.TYPES[this.type].isDismissableByDefault
```

The constants could contain more data about types and would make it easier to add new types, if the need ever rise.

I would understand if you considered it over-engineering 😄

PR #274, opened 2018-10-01
fix(oui-field): fix getFirstError return

Jisay:


packages/oui-field/src/field.controller.js
```diff
@@ -183,7 +183,7 @@ export default class FieldController {
             }
         }
 
-        return null;
```
jleveugle: Hey,

I don't think it's the right fix, we should return null instead of an empty object, what is the real issue ?

Thx

Jisay: I was writing a comment to explain that but here is the explanations :-) 
So when using an input with debounce in ngModelOptions the oui-field controller checks for first error to display through getMessagesOrder() call: https://github.com/ovh-ux/ovh-ui-angular/blob/3a5e2668a8d26b195ebce7d7c7a8fc4d7f6b975e/packages/oui-field/src/field.controller.js#L189-L192
But when getFirstError return null, an error occurs.
So an other solution is to patch line 191 this way:
```javascript
return Object.keys(this.getFirstError() || {}).sort((a, b) => MESSAGES_ORDER.indexOf(a) < MESSAGES_ORDER.indexOf(b));
```

jleveugle: Before using Object.keys, we should check if the parameter is a real object (by using typeof or another similar solution :) )

Imo, it's legit here to return null, the parent code should check if the return is not null, it's enought, nop?

AxelPeter: I like more too the idea that it's the one who's using the function that must check its return :)

PR #284, opened 2018-10-09
fix(oui-select-picker): fix selected value on init

AxelPeter:
* fix selectedValue on init
* fix `id` for accessibility
* update documentation

### DIFF
--- a/packages/oui-select-picker/src/select-picker.html
+++ b/packages/oui-select-picker/src/select-picker.html
@@ -1,8 +1,8 @@
 <input type="radio" class="oui-select-picker__input"
         name="{{:: $ctrl.name }}"
-        ng-attr-aria-labelledby="{{:: $ctrl.id }}-label"
-        ng-attr-aria-describedby="{{:: $ctrl.description && $ctrl.id + '-description' || undefined }}"
-        ng-attr-id="{{:: $ctrl.selectedValue ? $ctrl.id : undefined }}"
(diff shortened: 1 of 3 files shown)


packages/oui-select-picker/src/select-picker.html
```diff
@@ -1,8 +1,8 @@
 <input type="radio" class="oui-select-picker__input"
         name="{{:: $ctrl.name }}"
-        ng-attr-aria-labelledby="{{:: $ctrl.id }}-label"
-        ng-attr-aria-describedby="{{:: $ctrl.description && $ctrl.id + '-description' || undefined }}"
-        ng-attr-id="{{:: $ctrl.selectedValue ? $ctrl.id : undefined }}"
+        ng-attr-aria-labelledby="{{:: $ctrl.id }}Label"
```
FredericEspiau: It could be nice to document conventions for UI Angular somewhere :)

packages/oui-select-picker/src/select-picker.html
```diff
@@ -1,8 +1,8 @@
...
+        ng-attr-aria-describedby="{{:: $ctrl.description && $ctrl.id + 'Description' || undefined }}"
```
FredericEspiau: `ng-attr-aria-describedby="{{:: ($ctrl.description && $ctrl.id + 'Description') || undefined }}"` is more readable maybe :)

PR #291, opened 2018-10-12
feat(oui-select): allow items to be disabled

marie-j:

### DIFF
--- a/packages/oui-select/README.md
+++ b/packages/oui-select/README.md
@@ -68,6 +68,24 @@
 </oui-select>
 ```
 
+### Disabled Items
+
+```html:preview
+<oui-select name="country"
+    model="$ctrl.country"
+    data-title="Select a country"
+    placeholder="Select a country..."
+    items="$ctrl.countries"
+    disable-items="$ctrl.disableItems($item)"
+    required
+    match="name"
+    data-align="start">
+    <span ng-bind="$item.name"></span>
+</oui-select>
+```
+
+**Note**: For each `$item` in `items` array, `disable-item` will be called with current `$item` as an argument. If it returns true, `$item` will be disabled. 
+
 ### On Change
 
 **Note**: Model will not be refreshed until the `on-change` callback hasn't returned. If you want to access the new model inside the `on-change` callback you need to use the `modelValue` variable as below.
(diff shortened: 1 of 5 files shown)


packages/oui-select/src/select.directive.js
```diff
@@ -17,6 +17,7 @@ export default () => ({
         title: "@?",
         placeholder: "@?",
         items: "<",
+        disabledItems: "<",
```
AxelPeter: This property should be optional: `<?`
Don't forget to update the API table too :)

packages/oui-select/README.md
```diff
@@ -68,6 +68,39 @@
...
+
+#### Using Object Array
+
+```html:preview
+<oui-select name="country"
+    model="$ctrl.country"
+    data-title="Select a country"
+    placeholder="Select a country..."
+    items="$ctrl.countries"
+    disabled-items="['Algeria', 'Canada', 'France', 'United States']"
```
AxelPeter: I think that you should add a note under/above these examples about the fact that `disabled-items` need an array of values based on the match :)

Since `match` is optional, how will work this property in the case of the property is not defined ?
This should be documented too.

packages/oui-select/src/select.controller.js
```diff
@@ -29,6 +32,14 @@ export default class {
             matchElement.html("{{$select.selected}}");
         }
 
+        if (isArray(this.disabledItems)) {
+            this.items = angular.copy(this.items).map(item =>
```
AxelPeter: We should use another property like `this.sortedItems` to keep the value of `this.item`.
What will happened when `items` or `disabled-items` is updated after the init ?

packages/oui-select/src/select.controller.js
```diff
@@ -29,6 +32,14 @@ export default class {
             matchElement.html("{{$select.selected}}");
         }
 
+        if (isArray(this.disabledItems)) {
```
AxelPeter: This should be in `$onInit` instead :)

packages/oui-select/README.md
```diff
@@ -68,6 +68,22 @@
...
 
+### Disabled Items
+
+```html:preview
+<oui-select name="country"
+    model="$ctrl.country"
+    data-title="Select a country"
+    placeholder="Select a country..."
+    items="$ctrl.countries"
+    disable-items="$ctrl.disableItems($item)"
```
AxelPeter: What's behind `$ctrl.disableItems($item)` (without looking on specs, just by reading the docs) ? 
I think we should add a note or an example that explain how this function work (since it's a filter).

Also, don't you have a PR to link with `ovh-ui-kit-documentation` ?

packages/oui-select/src/select.controller.js
```diff
@@ -17,6 +17,7 @@ export default class {
     $onInit () {
         addBooleanParameter(this, "disabled");
         addBooleanParameter(this, "required");
+
```
AxelPeter: No need for this blank space :)

PR #296, opened 2018-10-22
feat(oui-datagrid): permit to access row index

marie-j:
### Description of the Change

Before: it was not possible to use row index value 
Now: Row index can be accessed with `$rowIndex`


packages/oui-datagrid/src/index.spec.js
```diff
@@ -1076,6 +1076,30 @@ describe("ouiDatagrid", () => {
...
+
+            const $firstRow = getRow(element, 0);
+
+            const actualCellHtml = getCell($firstRow, 1)
+                .children()
+                .children()
+                .html()
+                .trim();
+
+            expect(actualCellHtml).toBe("test: 0");
```
AxelPeter: Maybe we should test another indexes than 0 as a precaution.

PR #300, opened 2018-10-25
fix(oui-select-picker): prevent leaving blank space

marie-j:
### Description of the Change

If there was no `match` attribute set and a section was added, an empty section was created 
Also, if no `match` attribute, `select` displayed empty values

### DIFF
--- a/packages/oui-select-picker/src/select-picker.html
+++ b/packages/oui-select-picker/src/select-picker.html
@@ -28,10 +28,13 @@
             ng-bind=":: $ctrl.description"
             ng-if="$ctrl.description"></span>
     <span class="oui-select-picker__value-container oui-select-picker__section"
-        ng-if="$ctrl.values.length">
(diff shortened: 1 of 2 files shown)


packages/oui-select-picker/src/select-picker.html
```diff
@@ -28,20 +28,23 @@
             ng-bind=":: $ctrl.description"
             ng-if="$ctrl.description"></span>
     <span class="oui-select-picker__value-container oui-select-picker__section"
-        ng-if="$ctrl.values.length">
+            ng-if="$ctrl.values.length === 1 && $ctrl.match">
         <span class="oui-select-picker__value"
-                ng-bind=":: $ctrl.getFirstValueMatch($ctrl.match)"
-                ng-if="$ctrl.values.length === 1"></span>
+                ng-bind=":: $ctrl.getFirstValueMatch($ctrl.match)">
```
JeremyDec: No alignment ? (For lines 31 & 33)

PR #306, opened 2018-11-06
feat(oui-dropdown): allow click on link items

marie-j:
### Description of the Change

Before : `on-click` callback was not available for element with `state` or `href`
Now: `on-click` callback is available for all `oui-dropdown-item` (and therefore `oui-action-menu-item`)


packages/oui-dropdown/src/index.spec.js
```diff
@@ -212,6 +212,32 @@ describe("ouiDropdown", () => {
                 expect(link.attr("target")).toBe("_blank");
                 expect(link.attr("rel")).toBe("noopener");
             });
+
+            it("should call click callback", () => {
```
FredericEspiau: Getter better at this, nice 👍

PR #312, opened 2018-11-08
fix(oui-select-picker): fix transclude with component instead of span

AxelPeter:
UK-57

* Add transclude with `oui-select-picker-section` instead of targeting `span` 
* Add deprecated support

### DIFF
--- a/packages/oui-select-picker/src/select-picker.component.js
+++ b/packages/oui-select-picker/src/select-picker.component.js
@@ -5,7 +5,8 @@ export default {
     template,
     controller,
     transclude: {
-        sectionSlot: "?span"
+        sectionSlot: "?ouiSelectPickerSection",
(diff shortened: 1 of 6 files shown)


packages/oui-select-picker/src/select-picker.component.js
```diff
@@ -5,7 +5,8 @@ export default {
     template,
     controller,
     transclude: {
-        sectionSlot: "?span"
+        sectionSlot: "?ouiSelectPickerSection",
+        deprecatedSlot: "?span" // Deprecated: Replaced by "oui-select-picker-section"
```
FredericEspiau: Won't it break implementations ?

AxelPeter: No, it has been adapted

PR #313, opened 2018-11-12
Feat/dynamic step

Jisay:
In some condition steps can be empty. To avoid displaying the empty step, add possibility to display step dynamically with `ngIf`.

Add a new option: `position` which if provided will insert the step to given position in `ouiStepperCtrl.steps` array.

### DIFF
--- a/packages/oui-button/src/button.html
+++ b/packages/oui-button/src/button.html
@@ -16,7 +16,7 @@
 >
   <span aria-hidden="true" class="oui-icon oui-icon_circle oui-icon-chevron-left"
     ng-if="::$ctrl.variantNav === 'previous'"></span>
-  <span ng-transclude>{{::$ctrl.text}}</span>
+  <span ng-transclude>{{$ctrl.text}}</span>
   <span aria-hidden="true" class="oui-icon oui-icon_circle oui-icon-chevron-right"
     ng-if="::$ctrl.variantNav === 'next'"></span>
 </button>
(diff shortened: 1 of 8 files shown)


packages/oui-button/src/button.html
```diff
@@ -16,7 +16,7 @@
 >
   <span aria-hidden="true" class="oui-icon oui-icon_circle oui-icon-chevron-left"
     ng-if="::$ctrl.variantNav === 'previous'"></span>
-  <span ng-transclude>{{::$ctrl.text}}</span>
+  <span ng-transclude>{{$ctrl.text}}</span>
```
AxelPeter: Can you update the API table too, if you remove the one-time binding ?

packages/oui-stepper/src/step-form/step-form.html
```diff
@@ -45,7 +45,7 @@
             ng-if="$ctrl.navigation"
             ng-show="!$ctrl.loading && $ctrl.stepper.focused">
             <oui-button variant="primary" type="submit"
-                text="{{::$ctrl.submitText || ($ctrl.stepper.last ? $ctrl.translations.submitButtonLabel : $ctrl.translations.nextButtonLabel)}}">
+                text="{{$ctrl.submitText || ($ctrl.stepper.last ? $ctrl.translations.submitButtonLabel : $ctrl.translations.nextButtonLabel)}}">
```
AxelPeter: For this one too :)

PR #316, opened 2018-11-16
fix(oui-select-picker): default value

ganeshkumar1989:
Close MBP-34

### Requirements

When there are multiple select-pickers and when more than 1 select picker have multiple 'values' (values.length > 1), then the value attribute of multiple radio buttons created out of the select picker have undefined values i.e. the value attributes are not unique

### Description of the Change

To fix this issue, the first value of the values array is taken as the default value (selectedValue) for a select-picker so that the select pickers have unique values

### DIFF
--- a/packages/oui-select-picker/src/select-picker.controller.js
+++ b/packages/oui-select-picker/src/select-picker.controller.js
@@ -31,12 +31,10 @@ export default class SelectPickerController {
         }
 
         if (this.values) {
-            if (this.values.length === 1) {
-                this.selectedValue = this.values[0];
-            }
-
             if (this.model && find(this.values, this.model)) {
                 this.selectedValue = this.model;
+            } else {
+                this.selectedValue = this.values[0];
             }
         }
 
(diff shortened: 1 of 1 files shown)


packages/oui-select-picker/src/select-picker.controller.js
```diff
@@ -31,12 +31,10 @@ export default class SelectPickerController {
...
 
         if (this.values) {
-            if (this.values.length === 1) {
-                this.selectedValue = this.values[0];
-            }
-
             if (this.model && find(this.values, this.model)) {
                 this.selectedValue = this.model;
+            } else {
+                this.selectedValue = this.values[0];
```
JeremyDec:
```suggestion
                [this.selectedValue] = this.values;
```

ganeshkumar1989: Thanks for the suggestion. This has been incorporated

AxelPeter:
@ganeshkumar1989  Can you target your PR to `master` ?
Since it's a fix, I could release a patch version for your task.

ganeshkumar1989:
Please refer to https://github.com/ovh-ux/ovh-ui-angular/pull/317

PR #323, opened 2018-11-23
feat(oui-datagrid): add on-row-select event

euhmeuh:


packages/oui-datagrid/src/datagrid.directive.js
```diff
@@ -17,7 +17,8 @@ export default () => {
             rowsLoader: "&?",
             rowLoader: "&?",
             emptyPlaceholder: "@?",
-            onColumnsParametersChange: "&"
+            onColumnsParametersChange: "&",
+            onRowSelect: "&"
```
JeremyDec: Shouldn't this option be optonnal ?

packages/oui-datagrid/README.md
```diff
@@ -713,6 +717,7 @@ call `rows-loader` and then a `row-loader` call for each line.
 | `selectable-rows`                 | boolean   | <?        | no                  | `true`, `false`  | `false`      | if rows can be selected
 | `columns-parameters`              | array     | <?        | no                  | n/a              | `undefined`  | columns parameters (see below)
 | `on-columns-parameters-change`    | function  | &         | no                  | n/a              | n/a          | triggered on column parameter change when datagrid is customizable
+| `on-row-select`                   | function  | &         | no                  | n/a              | n/a          | triggered when a row is selected
```
JeremyDec: `&?` if the option is optionnal

AxelPeter: It's made on purpose. `&` call `angular.noop()` instead of `undefined` (with `&?`). 
It can be very useful inside template, and can avoid an if.

packages/oui-datagrid/README.md
```diff
@@ -123,6 +123,10 @@ Clicked row action 1: <span ng-if="$ctrl.action1Row">{{$ctrl.action1Row.lastName
 </oui-datagrid>
 ```
 
+<oui-message type="info" dismissable="false">
```
AxelPeter: Much better :)

packages/oui-datagrid/README.md
JeremyDec: Ok nice, it's good to know

FredericEspiau: It's lying to the users though, we could have a `mandatory` column that declares if a parameter is mandatory, independently from the binding

packages/oui-datagrid/README.md
```diff
@@ -94,7 +94,7 @@ Clicked row action 1: <span ng-if="$ctrl.action1Row">{{$ctrl.action1Row.lastName
 ### Selectable rows
 
 ```html:preview
-<oui-datagrid rows="$ctrl.data" page-size="5" selectable-rows>
+<oui-datagrid rows="$ctrl.data" page-size="5" selectable-rows on-row-select="$ctrl.onRowSelect($row, $rows)">
```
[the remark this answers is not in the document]
AxelPeter: I think we should keep a standard `on{Verb}` like our others events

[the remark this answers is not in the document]
FredericEspiau: Interesting question, I do like having the `target` before the `action`

Here are some conventions I've found:

- https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members
Recommends having this method named `ClickedRowHandler`. If it were triggered before the click, then the name would be `ClickingRowHandler`

- https://segment.com/academy/collecting-data/naming-conventions-for-clean-data/
Things that the event should be named `rowClicked` but doesn't say anything about event handler

- https://jaketrent.com/post/naming-event-handlers-react/
There you can find a convention for React, the naming of the event handler would then be `onRowClick`

- Bottom line
We could name events `['targetName', capitalize('actionName'), 'tenseSuffix'].join('')` :
`rowClicking` for before the event
`rowClicked` for after the event

We could name the handler `['on', capitalize('targetName'), capitalize('actionName'), 'tenseSuffix'].join('')` :
`onRowClicking`
`onRowClicked`

what do you think ?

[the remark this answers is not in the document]
AxelPeter: Ok for `on{TargetName}{ActioneName]. That's what is currently use for all our components and closer of JS events standards. But I think there is no need for a tenseSuffix here. 
It could be useful in another case, when a before/after avent is needed.

And also, I found that `onRowClick` is not really appropriate for this event. It could be interpreted as a click on all the row. We can keep `onRowSelect`, or something like `onRowSelectionClick`.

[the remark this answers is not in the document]
euhmeuh: I'm with Axel here, I think we don't have any use for the past tense yet. Plus, yes, the selection is definitely not a click, so `onRowSelect` seems good to me.

PR #324, opened 2018-11-27
feat(oui-file): add file component

AxelPeter:
UK-60

Related to
* https://github.com/ovh-ux/ovh-ui-kit/pull/312
* https://github.com/ovh-ux/ovh-ui-kit-documentation/pull/174

### DIFF
--- a/packages/oui-field/src/field.html
+++ b/packages/oui-field/src/field.html
@@ -23,7 +23,7 @@
         ng-if="$ctrl.isErrorVisible()">
         <span ng-messages="$ctrl.getFirstError()" role="alert">
             <span class="oui-icon oui-icon-error" aria-hidden="true"></span>
-            <span data-ng-repeat="errorKey in $ctrl.getMessagesOrder() track by $index"
+            <span data-ng-repeat="errorKey in $ctrl.getMessagesOrder()"
                   ng-message="{{ :: errorKey }}"
                   ng-bind="$ctrl.getErrorMessage(errorKey)">
             </span>
--- a/packages/oui-file/README.md
+++ b/packages/oui-file/README.md
@@ -105,7 +105,8 @@ angular.module("myModule", [
         dropAreaSelector: "select a file",
         fileSelector: "Select file",
         filesSelector: "Select file(s)...",
-        maxsizeError: "This file exceeds the size limit"
+        maxsizeError: "This file exceeds the size limit",
+        removeFile: "Remove file from selector"
     });
     ouiFileConfigurationProvider.setUnits([ // default units
         { size: 1000000000, suffix: "GB" },
--- a/packages/oui-file/README.md
+++ b/packages/oui-file/README.md
@@ -1,0 +1,117 @@
+# File
+
+<component-status cx-design="complete" ux="rc"></component-status>
+
+## Selector
+
+### Basic
+
+```html:preview
+<oui-file model="$ctrl.modelBasic"></oui-file>
+```
+
+### Placeholder
+
+```html:preview
+<oui-file model="$ctrl.modelPlaceholder" placeholder="Placeholder example"></oui-file>
+```
+
+### Disabled
+
+```html:preview
+<oui-file model="$ctrl.modelDisabled" disabled></oui-file>
+```
+
+### File restriction
+
+<oui-message type="info" dismissable="false">`maxsize` support form validators</oui-message>
+
+```html:preview
+<form novalidate name="fileForm">
+    <oui-field label="Upload file" help-text="image/jpeg, image/png, image/gif (max size 150 KB)" size="xl">
+        <oui-file name="fileUpload" model="$ctrl.modelField" 
+            accept="image/jpeg,image/png,image/gif" 
+            maxsize="150000" 
+            required>
+        </oui-file>
+    </oui-field>
+    <p><strong><code class="oui-doc-codespan">fileForm.$valid</code> value:</strong> {{fileForm.$valid | json}}</p>
+    <p><strong><code class="oui-doc-codespan">fileForm.$submitted</code> value:</strong> {{fileForm.$submitted | json}}</p>
+    <p><strong><code class="oui-doc-codespan">fileForm["fileUpload"].$error</code> value:</strong> {{fileForm["fileUpload"].$error | json}}</p>
+    <oui-button type="submit">Submit</oui-button>
+</form>
+```
+
+## Multiple files
+
+### Basic
+
+```html:preview
+<oui-file model="$ctrl.modelMultiple" maxsize="150000" multiple></oui-file>
+```
+
+### With preview
+
+<oui-message type="info" dismissable="false">Preview works only with `image/*` files.</oui-message>
+
+```html:preview
+<oui-file model="$ctrl.modelMultiplePreview" maxsize="150000" multiple preview></oui-file>
+```
+
+## Drag & Drop area
+
+### Basic
+
+```html:preview
+<oui-file model="$ctrl.modelDroparea" maxsize="150000" droparea></oui-file>
+```
+
+### With preview
+
+<oui-message type="info" dismissable="false">Preview works only with `image/*` files.</oui-message>
+
+```html:preview
+<oui-file model="$ctrl.modelDropareaPreview" maxsize="150000" droparea preview></oui-file>
+```
+
+## API
+
+| Attribute         | Type              | Binding   | One-time binding  | Values            | Default   | Description
+| ----              | ----              | ----      | ----              | ----              | ----      | ----
+| `model`           | array&lt;file&gt; | =         | no                | n/a               | n/a       | model bound to component
+| `id`              | string            | @?        | yes               | n/a               | n/a       | id attribute of form input
+| `name`            | string            | @?        | yes               | n/a               | n/a       | name attribute of form input
+| `placeholder`     | string            | @?        | yes               | n/a               | n/a       | placeholder text
+| `accept`          | string            | @?        | yes               | n/a               | n/a       | accept attribute of file input
+| `maxsize`         | number            | <?        | no                | n/a               | n/a       | maximum file size in byte
+| `disabled`        | boolean           | <?        | no                | `true`, `false`   | `false`   | disabled flag
+| `required`        | boolean           | <?        | no                | `true`, `false`   | `false`   | required flag
+| `multiple`        | boolean           | <?        | yes               | `true`, `false`   | `false`   | multiple flag
+| `droparea`        | boolean           | <?        | yes               | `true`, `false`   | `false`   | droparea flag
+| `preview`         | boolean           | <?        | yes               | `true`, `false`   | `false`   | preview flag
+| `on-select`       | function          | &         | no                | n/a               | n/a       | handler triggered when files are selected
+
+## Configuration
+
+The file component can be globally configured with a provider.
+
+```js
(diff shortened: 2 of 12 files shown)


packages/oui-field/src/field.html
```diff
@@ -23,7 +23,7 @@
         ng-if="$ctrl.isErrorVisible()">
         <span ng-messages="$ctrl.getFirstError()" role="alert">
             <span class="oui-icon oui-icon-error" aria-hidden="true"></span>
-            <span data-ng-repeat="errorKey in $ctrl.getMessagesOrder() track by $index"
+            <span data-ng-repeat="errorKey in $ctrl.getMessagesOrder()"
```
FredericEspiau: Why ?

packages/oui-file/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+```html:preview
+<oui-file model="$ctrl.modelDropareaPreview" maxsize="150000" droparea preview></oui-file>
+```
+
+## API
+
+| Attribute         | Type              | Binding   | One-time binding  | Values            | Default   | Description
+| ----              | ----              | ----      | ----              | ----              | ----      | ----
+| `model`           | array&lt;file&gt; | =         | no                | n/a               | n/a       | model bound to component
```
FredericEspiau: What's a file ? Binary data ? Doesn't it have properties ?

packages/oui-file/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+The file component can be globally configured with a provider.
+
+```js
+angular.module("myModule", [
+    "oui.file"
+]).config(ouiFileConfigurationProvider => {
+    ouiFileConfigurationProvider.setTranslations({ // default translations
+        attachmentsHeading: "Attachment(s)",
+        dropArea: "Attach document(s) by drap and drop or",
```
FredericEspiau: The `s` could be deduced from the presence or not of the `multiple` attribute :)

packages/oui-file/README.md
```diff
@@ -0,0 +1,117 @@
...
+## API
+
+| Attribute         | Type              | Binding   | One-time binding  | Values            | Default   | Description
+| ----              | ----              | ----      | ----              | ----              | ----      | ----
+| `model`           | array&lt;file&gt; | =         | no                | n/a               | n/a       | model bound to component
+| `id`              | string            | @?        | yes               | n/a               | n/a       | id attribute of form input
+| `name`            | string            | @?        | yes               | n/a               | n/a       | name attribute of form input
+| `placeholder`     | string            | @?        | yes               | n/a               | n/a       | placeholder text
+| `accept`          | string            | @?        | yes               | n/a               | n/a       | accept attribute of file input
+| `maxsize`         | number            | <?        | no                | n/a               | n/a       | maximum file size in byte
```
FredericEspiau: `on-select` but `maxsize` ? (`on-select` is separated by an hyphen but `maxsize` is all attached)

FredericEspiau: default max size ? Users could try to upload 1 Tb files and think that their browser is frozen because of us

FredericEspiau: wouldn't it make sense to be able to select a unit too ?

packages/oui-file/src/file.controller.js
```diff
@@ -0,0 +1,204 @@
...
+        if (this.preview && !file.errors && file.type && file.type.search(/^image\//) !== -1) {
+            file.loading = true;
+
+            file.reader = new this.$window.FileReader();
+            file.reader.readAsDataURL(file);
+            file.reader.onload = () => {
+                file.preview = `url("${file.reader.result}")`;
+
+                // Clean file object
+                delete file.loading;
```
FredericEspiau: Why do you delete all the properties ?

packages/oui-file/src/file.controller.js
```diff
@@ -0,0 +1,204 @@
...
+
+                this.$scope.$apply();
+            };
+        }
+
+        return file;
+    }
+
+    addFile (file) {
+        this.model = [this.checkFileValidity(file)];
```
FredericEspiau: Won't the `model` only contain last file ? Also, you still add the file even though it's invalid or empty ?

FredericEspiau: Just realized you have a `addFiles` method. I would have had only one method that can handle the case where their are one or more than one file

packages/oui-file/src/file.controller.js
```diff
@@ -0,0 +1,204 @@
...
+            });
+
+            this.onSelect({ modelValue: this.model });
+            this.$scope.$apply();
+        }
+    }
+
+    removeFile (file) {
+        if (angular.isArray(this.model)) {
+            remove(this.model, (item) => item === file);
```
FredericEspiau: So a `file` is a string ? Wouldn't it make more sense to have it being an object with a `path` property ?

```js
{
  path: '...'
  isReadingPreview: true,
  preview: [...]
}
```

packages/oui-file/src/file.controller.js
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    openFileSelector () {
+        // triggerHandler("click") don't work here
+        this.fileSelector[0].click();
+    }
+
+    static getFileName (file) {
+        // Remove file extension from name
+        const parts = file.name.split(".");
```
FredericEspiau:
```suggestion
        return file.name.replace(/\.[^/.]+$/, "");
```

FredericEspiau: But what if the file doesn't have an extension ?

FredericEspiau: I want to upload my `.gitignore` file

packages/oui-file/src/file.html
```diff
@@ -0,0 +1,132 @@
...
+<div class="oui-file-selector"
+    ng-if="::!$ctrl.multiple && !$ctrl.droparea"
+    ng-class="{
+        'oui-file-selector_error': $ctrl.form && $ctrl.form[$ctrl.name].$invalid && ($ctrl.form[$ctrl.name].$touched || $ctrl.form.$submitted)
+    }">
+
+    <!-- Fake Selector -->
+    <div class="oui-file-selector__input">
+        <button class="oui-file-selector__label" type="button"
+            ng-blur="$ctrl.setInputTouched()"
```
FredericEspiau: Set the input as touched when it's blured ?

packages/oui-file/src/file.html
```diff
@@ -0,0 +1,132 @@
...
+            <span class="oui-file-ellipsis"
+                ng-if="!!$ctrl.model[0]">
+                <span class="oui-file-ellipsis__name" ng-bind="$ctrl.constructor.getFileName($ctrl.model[0])"></span>
+                <span class="oui-file-ellipsis__extension" ng-bind="$ctrl.constructor.getFileExtension($ctrl.model[0])"></span>
+                <span class="oui-file-ellipsis__size" ng-bind="$ctrl.getFileSize($ctrl.model[0])"></span>
+            </span>
+        </button>
+        <button class="oui-file__remove" type="button"
+            ng-show="!!$ctrl.model[0]"
+            ng-click="$ctrl.resetFile()"
```
FredericEspiau: It looks more like `removeFile`

packages/oui-file/src/file.html
```diff
@@ -0,0 +1,132 @@
...
+            ng-disabled="$ctrl.disabled">
+            <span class="oui-icon oui-icon-close_thin"></span>
```
FredericEspiau: `aria-hidden`

FredericEspiau: Alternative text for screen readers

packages/oui-file/src/file.html
```diff
@@ -0,0 +1,132 @@
...
+                </span>
+                <span class="oui-file-attachments__error"
+                    ng-if="file.errors && file.errors.maxsize"
+                    ng-bind="::$ctrl.translations.maxsizeError">
+                </span>
+            </span>
+            <button class="oui-file__remove" type="button"
+                ng-click="$ctrl.removeFile(file)"
+                ng-disabled="$ctrl.disabled">
+                <span class="oui-icon oui-icon-close_thin"></span>
```
FredericEspiau: `aria-hidden` and alternative text

packages/oui-file/src/file.controller.js
AxelPeter: Read my comment

packages/oui-field/src/field.html
AxelPeter: It doesn't refresh the error, and displayed nothing

packages/oui-file/README.md
AxelPeter: https://developer.mozilla.org/fr/docs/Web/API/File

packages/oui-file/README.md
AxelPeter: Drop area is multiple only

packages/oui-file/README.md
AxelPeter: Like we can do with the current input file ?

`on-select` -> Keep our way on naming event
`maxsize` -> same as `maxlength` attribute

packages/oui-file/src/file.html
AxelPeter: Form validation

packages/oui-file/src/file.html
AxelPeter: No, see the method

packages/oui-file/src/file.html
AxelPeter: Why hidden, since it's the close button icon and its only content ?
I'll check for the `aria-label`.

packages/oui-file/src/file.html
AxelPeter: Same as my upper comment

packages/oui-file/src/file.controller.js
AxelPeter: In your case, `gitignore` is the extension. And the filename is ` `.
But I thought about one thing. It will not work with a `.tag.gz` file. It'll only take `.gz` has an extension. I'll check to modify that

packages/oui-file/src/file.controller.js
AxelPeter: No, a file is a file object. What make you think it's a string ?

packages/oui-file/src/file.controller.js
AxelPeter: I've removed the "delete" lines. Could be still useful.

packages/oui-file/src/file.html
FredericEspiau: I've got no away to know if it's a close button or a delete my account button or a like on facebook button if I'm on a screen reader

packages/oui-file/src/file.controller.js
FredericEspiau: Did you find a solution ?

packages/oui-file/src/file.html
AxelPeter: done

packages/oui-file/src/file.html
AxelPeter: done

packages/oui-file/src/file.controller.js
AxelPeter: It seems there is no real solution, even from the MIME. So it'll stay as is, since it's real purpose is for ellipsis. But I'll need to cover files like `test` or `.test`.

PR #327, opened 2018-12-04
refactor(oui-select): use ui-select in a proper way

AxelPeter:
UK-64

Related to
* https://github.com/ovh-ux/ovh-ui-kit/pull/318


packages/oui-select/src/templates/match-multiple.html
```diff
@@ -0,0 +1,16 @@
+<span class="ui-select-match">
+    <span ng-repeat="$item in $select.selected track by $index">
+        <span class="ui-select-match-item" type="button"
+            ng-disabled="$select.disabled"
+            ng-click="$selectMultiple.activeMatchIndex = $index;"
+            ng-class="{
+                'btn-primary': $selectMultiple.activeMatchIndex === $index,
```
antleblanc: I'd like to know why in that case you didn't prefer using `oui-button_primary` class?

Are there specific reason? Thanks a lots for your feedback.

AxelPeter: Yes, it's normal for now for all the parts about multiple choices. This will be done in a further ticket :)

PR #337, opened 2018-12-26
fix(oui-header-tabs): remove active link if child state is active

JeremyDec:


JeremyDec:
Closed because of drawbacks behavior unwanted

PR #338, opened 2018-12-27
fix(oui-select): nested property support for match api

radireddy:
support nested property like "region.name" on match input in oui-select component

Closes #MANAGER-2056

### Requirements

## Title of the Pull Requests 
support nested property like "region.name" on match input in oui-select component

### Description of the Change
oui-select-picker uses oui-select. This issue happens If match property has nested properties like 'region.name.' The reason is that of a bug in match.html used by oui-select. It does not take the nested property into consideration.

```
 <span class="ui-select-match-text"
            ng-hide="$select.isEmpty()"
            ng-bind-html="$select.selected[$ctrl.match] || $select.selected">
  </span>
```
in the above code, we straight away getting a property on the object. But the supplied property is a nested property.

Solution: used loadash get that supports nested property.

### Possible Drawbacks
none

### Applicable Issues
MANAGER-2056

### DIFF
--- a/packages/oui-select/src/templates/match.html
+++ b/packages/oui-select/src/templates/match.html
@@ -15,7 +15,7 @@
         </span>
         <span class="ui-select-match-text"
             ng-hide="$select.isEmpty()"
-            ng-bind-html="$select.selected[$ctrl.match] || $select.selected">
+            ng-bind-html="$ctrl.getPropertyValue($select.selected) || $select.selected">
         </span>
         <span class="oui-icon oui-icon-chevron-down"></span>
     </span>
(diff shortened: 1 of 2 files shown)


packages/oui-select/src/templates/match.html
```diff
@@ -15,7 +15,7 @@
         </span>
         <span class="ui-select-match-text"
             ng-hide="$select.isEmpty()"
-            ng-bind-html="$select.selected[$ctrl.match] || $select.selected">
+            ng-bind-html="$ctrl.getPropertyValue($select.selected) || $select.selected">
```
AxelPeter: Since `getPropertyValue` can return a default value with `get`, maybe we should go with;

```
ng-bind-html="$ctrl.getPropertyValue($select.selected)"
```

 and

```
    getPropertyValue (item) {
        return get(item, this.match, item);
    }
```

What do you think ?

PR #339, opened 2019-01-03
fix(oui-select): fix match template of oui-ui-select theme

Jisay:
Allow to use `oui-ui-select` theme directly with ui-select directive. This allow to customize the selected item.
This checks if match attribute is set and if not set, transclude the ui-select-match directive content.


AxelPeter:
Seen together, it'll be done by overriding the template in the manager.

PR #342, opened 2019-01-08
fix(oui-slideshow): fix onDismiss callback

AxelPeter:
Fix UK-84:
`ReferenceError: "event is not defined"` console error


AxelPeter:
Since the binding is `"onDismiss": "&"`, there is no need to check if `undefined`.

PR #353, opened 2019-01-25
fix(oui-file): fix input hidden error from ng 1.7.6

AxelPeter:
Since [AngularJS 1.7.6](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#176-gravity-manipulation-2019-01-17), there is an error when using `input[hidden]` in oui-file.

To avoid this, the `input[hidden]` has been changed to `input[text]`, hidden using CSS.

### DIFF
--- a/packages/oui-file/src/file.html
+++ b/packages/oui-file/src/file.html
@@ -1,18 +1,21 @@
 <!-- Controls -->
-<input type="hidden"
+<!-- Using [hidden] cause errors with ng 1.7.6, -->
+<!-- We use [text] hidden by css to avoid it -->
+<input class="oui-file__input oui-file__input_hidden" type="text"
+    autocomplete="off"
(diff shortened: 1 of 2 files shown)


antleblanc:
Shouldn't be handled in a the next version?

Both `devDependencies` and `peerDependencies` are targeting `>=1.6.x`.

https://github.com/ovh-ux/ovh-ui-angular/blob/master/package.json#L44
https://github.com/ovh-ux/ovh-ui-angular/blob/master/package.json#L51

AxelPeter:
@antleblanc Yes, and ? It's compatible with 1.6 and 1.7.
I don't see what you mean by your comment.

antleblanc:
In order to provide support to the `v1.7.6` of AngularJS and `>=`, you may have to reflect these changes on both `devDependencies` and `peerDependencies` as well, isn't it?

PR #365, opened 2019-02-25
feat(oui-popover): add scope attribute for template

AxelPeter:
* Add the possibility to extend the isolated scope create for the popover's template.
* Update documentation
* Update unit tests

### DIFF
--- a/packages/oui-popover/src/popover.directive.js
+++ b/packages/oui-popover/src/popover.directive.js
@@ -9,7 +9,7 @@ export default () => {
             text: "@ouiPopover",
             title: "@?",
             placement: "@?ouiPopoverPlacement",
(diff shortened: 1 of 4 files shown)


packages/oui-popover/src/popover.directive.js
```diff
@@ -9,6 +9,7 @@ export default () => {
             text: "@ouiPopover",
             title: "@?",
             placement: "@?ouiPopoverPlacement",
+            scope: "=?ouiPopoverScope",
```
FredericEspiau: Why not `<` ?

AxelPeter: Yes, no utility to have a two way here. Updated

PR #372, opened 2019-03-19
feat(oui-collapsible): add handler on toggle open-close event

svandecappelle:
### Description of the Change

On header click this now dispatch the toggle event to a function passed in attributes with true/false parameter that defines if the collapsible shows his contents or not.

### Benefits

Know the state of the collapsible on change

### DIFF
--- a/packages/oui-collapsible/src/collapsible.controller.js
+++ b/packages/oui-collapsible/src/collapsible.controller.js
@@ -38,5 +38,6 @@ export default class {
 
     toggle () {
         this.expanded = !this.expanded;
+        this.onToggle({ expanded: this.expanded });
     }
 }
--- a/packages/oui-collapsible/src/index.spec.js
+++ b/packages/oui-collapsible/src/index.spec.js
@@ -75,5 +75,23 @@ describe("ouiCollapsible", () => {
             const contentEl = bodyEl.querySelector(".custom-content");
             expect(contentEl).toBeTruthy();
         });
+
+        it("should call ontoggle when defined into attributes", () => {
+            const onToggle = jasmine.createSpy("onToggle");
+            const element = TestUtils.compileTemplate(`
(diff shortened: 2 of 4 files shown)


packages/oui-collapsible/src/collapsible.html
```diff
@@ -2,7 +2,7 @@
     type="button"
     ng-click="$ctrl.toggle()"
     aria-label="{{::$ctrl.ariaLabel}}"
-    aria-expanded="{{$ctrl.expanded}}"
+    aria-expanded="{{$ctrl.expanded}}"
```
AxelPeter: No need for this extra spaces

packages/oui-collapsible/src/index.spec.js
```diff
@@ -75,5 +75,19 @@ describe("ouiCollapsible", () => {
...
+            const element = TestUtils.compileTemplate(`
+                <oui-collapsible heading="Title" aria-label="Action" on-toggle="$ctrl.onToggle(expanded)"></oui-collapsible>`, {
+                onToggle
+            });
+
+            const headerEl = angular.element(getHeaderElement(element));
+
+            // Expand
+            headerEl.triggerHandler("click");
+            expect(onToggle).toHaveBeenCalledWith(true);
```
AxelPeter: I think we should add an `expect` when expanded is false :)

packages/oui-collapsible/src/collapsible.controller.js
```diff
@@ -38,5 +38,7 @@ export default class {
 
     toggle () {
         this.expanded = !this.expanded;
+        const expanded = this.expanded;
+        this.onToggle({ expanded });
```
AxelPeter:
```suggestion
        this.onToggle({ expanded: this.expanded });
```

I think we could go with this suggestion. It's shorter and still readable ;)

AxelPeter:
Here's the API table that need to be upated: https://github.com/ovh-ux/ovh-ui-angular/blob/master/packages/oui-collapsible/README.md, and we're good :)

PR #378, opened 2019-03-28
fix(oui-select-picker): update when model changes

cbourgois:
### Description of the Change

c1f51a0 — fix(oui-select-picker): update when model changes

/cc @jleveugle @frenautvh @antleblanc @marie-j @AxelPeter

### DIFF
--- a/packages/oui-select-picker/src/select-picker.controller.js
+++ b/packages/oui-select-picker/src/select-picker.controller.js
@@ -38,6 +38,12 @@ export default class SelectPickerController {
             }
         }
 
+        this.$scope.$watch(() => this.model, (newValue) => {
+            if (this.model && find(this.values, newValue)) {
+                this.selectedValue = newValue;
+            }
+        });
+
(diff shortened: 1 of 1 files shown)


packages/oui-select-picker/src/select-picker.controller.js
```diff
@@ -38,6 +38,12 @@ export default class SelectPickerController {
             }
         }
 
+        this.$scope.$watch("$ctrl.model", (newValue) => {
```
JeremyDec: This cannot be made through `$onChanges` hook ?

JeremyDec: This cannot be made through `$onChanges` hook ?

cbourgois: No it doesn't work.

AxelPeter: It's because of the binding of the `model` (=) :)

Can you prefer using a function instead ?
```suggestion
        this.$scope.$watch(() => this.model, (newValue) => {
```

PR #379, opened 2019-04-02
fix(oui-stepper): conditional step navigation

radireddy:
Custome validation is stepper component

Use case: Execute costume validation on submitting a stepper and do not go to the next step if the validation fails.

The only way to achieve this is to use the valid method of the stepper component. I have added my custom validation in a valid method but this method gets executed many many times which is not very efficient and taking longer time. I want to execute (only once) complex validation on form submit and go to the next step only if the validation succeeds. 

<oui-step-form data-header="Size the nodes"
             data-loading="$ctrl.supportedFlavors.loading && $ctrl.quota.loading"
             data-on-focus="$ctrl.initNodes()"
             data-valid="$ctrl.isNodesValid()"
             data-on-submit="$ctrl.submitNodesInformation()">
             <div data-ng-include="'adp/deploy/nodes.html'"></div>
  </oui-step-form>

Current behavior
a valid method of oui-step-form gets fired many many times even though the stepper is not focused. For example, if a stepper has 5 steps and the last step has a valid method. This method gets executed as soon as the stepper is loaded and get executed many times, more than 100, based on the complexity of the stepper form.
In the above example, isNodesValid() method gets called many times even before reaching this step. Since I have a complex validation in isNodesValid() method, it's taking too long. 

Expected behavior
valid method must be executed only on clicking the submit button for that step.

### Requirements

## Title of the Pull Requests 
Conditional step navigation

### Description of the Change
Go to the next step in stepper component only if on submit method return true or nothing. Stop step navigation if this method returns false.

### DIFF
--- a/packages/oui-stepper/src/step-form/step-form.controller.js
+++ b/packages/oui-stepper/src/step-form/step-form.controller.js
@@ -58,9 +58,8 @@ export default class StepFormController {
     }
 
     onFormSubmit (form) {
+        this.onSubmit({ form });
         if (form.$valid && this.valid) {
-            this.onSubmit({ form });
-
             // Focus next step
             this.stepperCtrl.addForm(form, this.stepper.index);
         }


packages/oui-stepper/src/step-form/step-form.controller.js
```diff
@@ -59,10 +60,11 @@ export default class StepFormController {
...
         if (form.$valid && this.valid) {
-            this.onSubmit({ form });
-
-            // Focus next step
-            this.stepperCtrl.addForm(form, this.stepper.index);
+            const gotToNextStep = this.onSubmit({ form });
+            if (isNil(gotToNextStep) || gotToNextStep === true) {
+                // Focus next step
+                this.stepperCtrl.addForm(form, this.stepper.index);
+            }
```
euhmeuh: We could just do that I guess:
```javascript
if (form.$valid && this.valid && this.onSubmit({ form })) {
    this.stepperCtrl.addForm(form, this.stepper.index);
}
```

AxelPeter: I'm really not a fan of using a callback of a callback, to get a value we can get using `valid` attribute.
Can you provide more informations about  your use case (with code example) in the description of this PR ?

A thing we could still do here, is to call `this.onSubmit({ form });` outside the condition, to allow access of the `form` value even when the form is not valid.

AxelPeter:
@radireddy Is something like this ticket could help you with your use case (UK-89) ?

packages/oui-stepper/src/step-form/step-form.controller.js
radireddy: @euhmeuh, we can do that because in all the places where stepper is used, the onSubmit method does not return anything but still need to go to the next step. The condition to go to the next step is if nothing is returned or true is returned.

radireddy: @AxelPeter , i have added detail description with my usecase and code sample.

radireddy: @AxelPeter , I undo my changes and moved onSubmit call before if statement. This will be sufficient for my use case and looks very neat as well.

radireddy:
@AxelPeter, As you suggested, I undo my changes and moved onSubmit call before if statement. This will be sufficient for my use case and looks very neat as well. Can you prod this now?

AxelPeter:
@radireddy This will be released with the next minor version (v2.27.0), when all the others PRs are resolved (ui-angular and ui-kit) :)

PR #381, opened 2019-04-04
feat(oui-stepper): add new attribute (editable)

jleveugle:
By default, after validating a step, user can edit it and update his choice.

```html
<oui-step-form header="Simple step" description="This is a description">
```

we want to let our developpers decide when this update is possible. Today, user can do it as he wishes, this can be problematic in case of loading for example.

Example:
```html
<oui-step-form header="Simple step" description="This is a description" editable="false">
```

will result by removing the link: "Modifying this step"


jleveugle:
> Need a unit test, and we're all good :)

done ;)

PR #389, opened 2019-04-15
fix(oui-datagrid): prevent pagination change from scrolling the view

marie-j:
### Description of the Change

Prevent automatic scrolling on pagination change 

### Benefits

Avoid unexpected side effects with component integration 

### Related 

https://github.com/ovh-ux/ovh-ui-kit-documentation/pull/192

### DIFF
--- a/packages/oui-datagrid/src/datagrid.controller.js
+++ b/packages/oui-datagrid/src/datagrid.controller.js
@@ -246,25 +246,23 @@ export default class DatagridController {
         this.refreshData(() => {
             this.paging.setOffset(1);
             this.paging.setCriteria(criteria);
-        }, false, false);
+        }, false);
     }
 
-    onPaginationChange ($event) {
+    onPaginationChange ({ offset, pageSize }) {
         this.refreshData(() => {
-            this.paging.setOffset($event.offset);
-            this.paging.setPageSize($event.pageSize);
-        }, true, true);
+            this.paging.setOffset(offset);
+            this.paging.setPageSize(pageSize);
+            this.onPageChange({
+                $pagination: {
+                    offset,
+                    pageSize
+                }
+            });
+        }, true);
     }
 
-    scrollToTop () {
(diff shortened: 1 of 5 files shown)


packages/oui-datagrid/src/datagrid.controller.js
```diff
@@ -246,25 +246,18 @@ export default class DatagridController {
...
 
-    onPaginationChange ($event) {
+    onPaginationChange ({ offset, pageSize }) {
         this.refreshData(() => {
-            this.paging.setOffset($event.offset);
-            this.paging.setPageSize($event.pageSize);
-        }, true, true);
+            this.paging.setOffset(offset);
+            this.paging.setPageSize(pageSize);
+            this.$scope.$emit("oui-datagrid.onPaginationChange", { offset, pageSize });
```
AxelPeter: Do you use this in your code ? I don't see a `$on('oui-datagrid.onPaginationChange')` here.
Why not callbacks `onPaginationChange` and `onDataRefresh` ?

FredericEspiau: I guess this is for users, it can be used outside the Component

marie-j: Indeed, this allows users to have custom behaviour on pagination change if they want (like using `scrollIntoView` if they want to :)

AxelPeter: This doesn't answer my last question above :)

Why prefer using `$emit` instead of callbacks ?
We could achieve exactly the same thing with callbacks too.

For me, it's more web-components friendly callbacks than emit/broadcast.

FredericEspiau: Wouldn't a callback for each event be worse than events ?

marie-j: @FredericEspiau are you talking in terms of performances ? 
It's true that for consistency a callback would be better

PR #390, opened 2019-04-16
feat(oui-password): validation trigger

ganeshkumar1989:
Trigger for validation in cases where the validation depends on other controls. This can be used, for example in confirm password field, where the filed depends on the actual password

Staging: https://ovh_ui_kit_documentation_new-feat_password_validation_trigger.uxci.ovh

### DIFF
--- a/packages/oui-password/README.md
+++ b/packages/oui-password/README.md
@@ -62,6 +62,32 @@
 </form>
 ```
 
+#### Confirm Validation
+
+Useful when you have a password field and a confirm password field. The confirm property takes the expression, the value of which should match with it's model value.
+
+```html:preview
+<form name="form3" novalidate>
+    <oui-field label="Password" size="xl">
+        <oui-password model="$ctrl.modelTriggerValidation">
+            <oui-password-rule validator="$ctrl.checkPasswordLength(modelValue)">
+                Must contain between 8 and 30 characters
+            </oui-password-rule>
+            <oui-password-rule pattern="[0-9]+">
+                Have at least one number
+            </oui-password-rule>
+            <oui-password-rule pattern="[A-Z]+">
+                Have at least capital letter
+            </oui-password-rule>
+        </oui-password>
+    </oui-field>
+    <oui-field label="Confirm Password" size="xl">
+        <oui-password model="$ctrl.modelTriggerValidationConfirmPassword" confirm="$ctrl.modelTriggerValidation">
+        </oui-password>
+    </oui-field>
+</form>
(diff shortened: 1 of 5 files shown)


packages/oui-password/README.md
```diff
@@ -62,6 +62,35 @@
...
+                Have at least one number
+            </oui-password-rule>
+            <oui-password-rule pattern="[A-Z]+">
+                Have at least capital letter
+            </oui-password-rule>
+        </oui-password>
+    </oui-field>
+    <oui-field label="Confirm Password" size="xl">
+        <oui-password model="$ctrl.modelTriggerValidationConfirmPassword" validation-trigger="$ctrl.modelTriggerValidation">
+            <oui-password-rule validator="$ctrl.checkPasswordMatch(modelValue)">
```
AxelPeter: I think we could go simpler with a `confirm` or `compare` attribute, since it will be its main use.

```
<oui-password model="$ctrl.model1"></oui-password>
<oui-password model="$ctrl.model2" confirm="$ctrl.model1"></oui-password>
```

The binding could be '=?', and watched if the attribute is present.
This will set `$setValidaty('confirm', value)` when changed and could trigger an error in the `oui-field` (https://github.com/ovh-ux/ovh-ui-angular/blob/master/packages/oui-field/src/field.provider.js#L7).

This seems easier to understand and to use, since the logic is already done and oriented to confirm another password :)

packages/oui-password/src/rule/password-rule.controller.js
```diff
@@ -31,8 +31,8 @@ export default class {
         );
 
         this.$scope.$watch(
-            () => this.password.model,
-            (value) => this.setValidity(value)
+            () => `${this.password.model}${this.password.validationTrigger}`,
```
AxelPeter: I think we could keep it like it was, with a simpler attribute ;)

packages/oui-password/README.md
ganeshkumar1989: This is a really good idea @AxelPeter 
This is  done

packages/oui-password/src/rule/password-rule.controller.js
ganeshkumar1989: This has be undone

ganeshkumar1989:
> Seems better to use :)
> Could you please add a unit test and remove the dist files from this commit ?
> Thanks !

Thanks a lot @AxelPeter . The unit test cases have been added and the dist files have been removed from the commit.

PR #391, opened 2019-04-17
fix(oui-stepper): use ng-if instead of ng-show for step form loading

marie-j:
### Description of the Change

Prevent the content from being transcluded while the loading isn't finish

### Benefits

Avoid troubles with transcluded elements needing asynchronous data

### DIFF
--- a/packages/oui-stepper/src/step-form/step-form.controller.js
+++ b/packages/oui-stepper/src/step-form/step-form.controller.js
@@ -15,6 +15,7 @@ export default class StepFormController {
         addBooleanParameter(this, "disabled");
         addBooleanParameter(this, "editable");
         addBooleanParameter(this, "skippable");
+        addBooleanParameter(this, "loading");
 
         // Add default name
         addDefaultParameter(this, "name", `ouiStepForm${this.$scope.$id}`);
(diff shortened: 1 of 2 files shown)


AxelPeter:
What's the troubles ? This isn't explained in the description.

We had a fix on the modal because we used a `ng-if` instead of `ng-show` for the exact same thing.
And now we do the reverse for the stepper.

marie-j:
> What's the troubles ? This isn't explained in the description.
> 
> We had a fix on the modal because we used a `ng-if` instead of `ng-show` for the exact same thing.
> And now we do the reverse for the stepper.

The issue I'm facing is that when you transclude a component that needs asynchronous data 
```
<oui-step-form loading="$ctrl.loading">
 <component binding="$ctrl.asynchronousData">
</oui-step-form>
````
with associated controller 
```
this.loading = true
this.getAsynchronousData()
 .finally(() => this.loading = false)
```
The `binding` will not be resolved correctly , we need to had `ng-if="$ctrl.loading"` on `component` and a spinner for loading state which imo duplicates what the `loading` attribute is here for

**Comment by AxelPeter** [comment]
Thanks for your update. 
But with your example, I don't think the problem comes from the component.

The problem we could have using a `ng-if` here is:
```html
<oui-step-form loading="$ctrl.loading">
 <component binding="$ctrl.getAsynchronousData()">
</oui-step-form>
```
`$ctrl.getAsynchronousData()` will never be called in this way.
And that's precisely why it was fixed with the modal (https://github.com/ovh-ux/ovh-ui-angular/commit/d7f863572ff0a6aa172955942d96df24f60bc138)

PR #393, opened 2019-04-18
feat(oui-stepper): add ability to set open step

marie-j:
### Description of the Change

Add `currentIndex` as binding to be able to set open state

### Benefits

To be able to switch between steps without user action

### DIFF
--- a/packages/oui-stepper/src/stepper.component.js
+++ b/packages/oui-stepper/src/stepper.component.js
@@ -6,7 +6,8 @@ export default {
         name: "@",
         id: "@",
         onInit: "&",
-        onFinish: "&"
+        onFinish: "&",
+        currentIndex: "=?"
     },
     controller,
     template,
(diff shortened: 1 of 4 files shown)


packages/oui-stepper/src/stepper.component.js
```diff
@@ -6,7 +6,8 @@ export default {
         name: "@",
         id: "@",
         onInit: "&",
-        onFinish: "&"
+        onFinish: "&",
+        currentIndex: "=?"
```
AxelPeter: Why not "<?" using `$onChanges` ?
We don't really need two way here :)

packages/oui-stepper/src/stepper.controller.js
```diff
@@ -21,6 +24,11 @@ export default class {
                 .removeAttr("name")
                 .addClass("oui-stepper")
         );
+
+        this.$scope.$watch(
```
Jisay: You can avoid using `$scope.$watch` by using `get`/`set` of es6 :-)

marie-j: I tried to but I get `RangeError: Maximum call stack size exceeded` (I suppose due to the fact that `currentIndex` is needed to be set elsewhere)

PR #398, opened 2019-05-07
feat(oui-tabs): allow to select active tab item

cbourgois:
### Description of the Change

d06498d — feat(oui-tabs): allow to select active tab item

### DIFF
--- a/packages/oui-tabs/src/tabs.controller.js
+++ b/packages/oui-tabs/src/tabs.controller.js
@@ -1,3 +1,5 @@
+import find from "lodash/find";
+
 export default class {
     constructor ($element, $timeout) {
         "ngInject";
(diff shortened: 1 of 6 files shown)


packages/oui-tabs/src/tabs.controller.js
```diff
@@ -50,5 +47,9 @@ export default class {
                 .addClass("oui-tabs")
                 .removeAttr("aria-label")
         );
+
+        if (this.items.length > 0) {
+            this.setActiveTab(find(this.items, tabItem => tabItem.id === this.model) || this.items[0]);
```
marie-j:
```suggestion
            this.setActiveTab(find(this.items, {id: this.model}) || this.items[0]);
```
as you are using lodash 😉

PR #402, opened 2019-05-15
feat(oui-stepper): emit event on form submit

marie-j:
### Benefits

Allow to associate global action on step form submission

### DIFF
--- a/packages/oui-stepper/src/step-form/step-form.controller.js
+++ b/packages/oui-stepper/src/step-form/step-form.controller.js
@@ -64,6 +64,7 @@ export default class StepFormController {
     }
 
     onFormSubmit (form) {
+        this.$scope.$emit("oui-step-form.submit", { form });
         this.onSubmit({ form });
         if (form.$valid && this.valid) {
             // Focus next step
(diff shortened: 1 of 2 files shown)


packages/oui-stepper/src/step-form/step-form.controller.js
```diff
@@ -64,6 +64,7 @@ export default class StepFormController {
     }
 
     onFormSubmit (form) {
+        this.$scope.$emit("oui-step-form.submit", { form });
```
antleblanc: Needed even if form is `$invalid`?

AxelPeter:
Why this need to be in the UI Kit ?

This can already be done with `on-submit="$ctrl.onSubmit(form)"`.
And $emit in your project with your callback. ;)

marie-j:
@AxelPeter I want to apply the same action (here tracking) on each steps submission. 
If I use the `onSubmit` callback I have to add it for each step form manually and also will have to add it for further added step-forms. By adding an event and listening to it in the hosts (if needed of course) we don't have add it manually each time

packages/oui-stepper/src/step-form/step-form.controller.js
marie-j: Yes it is

PR #406, opened 2019-05-28
feat(oui-form-actions): allow submit button to be disabled

marie-j:
### Description of the Change

Add possibility to pass a `disabled` parameter to disable submit button


AxelPeter:
@marie-j Need to be migrated to v3 too ;)

PR #409, opened 2019-06-10
fix(oui-dropdown): prevent popper from changing the placement

marie-j:
## Prevent popper from changing the placement

Prevent Popper from opening dropdown another way that specified


marie-j:
> What about the others components that use Popper.js ?

@AxelPeter The only place I found this option is used was in the `oui-popover` but I wonder if we want the same behaviour e.g : if the popover is supposed to open on the right it will open on the left on mobile to keep it into the viewport whereas without the option it will keep opening on the right which will imply horizontal scrolling, not sure it's a wanted behaviour

PR #411, opened 2019-06-24
Feat/error display

Jisay:
Proposal for ovh-ux/ovh-ui-kit#458.
We have to discuss about name of attribute and name of global configuration into provider.


packages/oui-field/src/field.component.js
```diff
@@ -7,7 +7,8 @@ export default {
         helpText: "@?",
         size: "@?",
         errorMessages: "<?",
-        labelPopover: "@?"
+        labelPopover: "@?",
+        errorDisplay: "&?"
```
JeremyDec: You don't need to specify the optional option :
```suggestion
        errorDisplay: "&"
```

AxelPeter: A question here could be: Do we need a callback in this case, instead of just a boolean attribute ? 

Currently, the errors are displayed when the field is `$dirty && $invalid` and blurred. But the error disappears when the field gets back the focus (eg. https://ovh-ux.github.io/ovh-ui-kit/?path=/story/components-field--input).

Maybe we could just simplify this part by removing the "focus" part. To be discussed :)

PR #413, opened 2019-06-26
Fix/one time binding

Jisay:
## oui-field

As fields labels can be dynamic remove one-time binding on labels and label-popover attributes.

## oui-stepper

As step can be dynamic too, remove one-time binding on header and description attributes.


Jisay:
Sure : ovh-ux/ovh-ui-kit#459 :-)

PR #424, opened 2019-08-27
fix(oui-datagrid): allow no property on oui-column

marie-j:
## Allow no property on oui-column


AxelPeter:
@marie-j Has this been ported to v3 ?

PR #427, opened 2019-09-09
fix(oui-dropdown): remove one time binding href

JeremyDec:
### Description of the Change
Remove the one time binding property to `href`attribute, to allow you to make dynamic url params on href


AxelPeter:
Should be ported to v3 too ;)

JeremyDec:
Yes, will do it today :)

PR #429, opened 2019-09-12
feat(oui-navbar): allow to pass link as menu header

marie-j:
### Description of the Change

Allow menu header to be a link


packages/oui-navbar/src/menu/navbar-menu.component.js
```diff
@@ -10,6 +10,7 @@ export default {
         headerBreadcrumb: "@?",
         headerClass: "@?",
         headerTitle: "@?",
+        headerLink: "@?",
```
AxelPeter: To keep consistency with all attributes about creating links, we should rename it `headerHref` :)

PR #434, opened 2019-10-11
fix(oui-field): checkAllErrors avoid undefined value

frenautvh:


AxelPeter:
@frenautvh Don't forget to port it on v3 ;)

PR #436, opened 2019-10-18
fix(oui-autocomplete): check focus before opening the autocompletion

jleveugle:


ovh-cds:
CDS Report yarn-build#1658.0 ✘  
* build & test
  * build ✔ 
  * test ✘

packages/oui-autocomplete/src/autocomplete.controller.js
```diff
@@ -232,6 +234,12 @@ export default class {
                 .attr("autocomplete", "off")
                 .attr("list", this.id)
                 .one("focus", () => this.createDatalist()) // One time bind to create the popper helper
+                .on("focus", () => {
```
FredericEspiau: Can't you use `ngFocus` instead ? I don't know what it better 😄

jleveugle: I keep the same logic, see the line before and after my update ;)

PR #446, opened 2019-12-23
fix(oui-stepper): stepper will be updated properly on step focus now

FredericEspiau:


FredericEspiau:
Not relevant, a better fix has been found :)

PR #451, opened 2020-01-06
fix(oui-calendar): dynamic changes for min and max date

AxelPeter:
ref: UK-123

### DIFF
--- a/packages/oui-calendar/src/calendar.controller.js
+++ b/packages/oui-calendar/src/calendar.controller.js
@@ -101,7 +101,6 @@ export default class {
     }
 
     $onChanges ({ minDate, maxDate }) {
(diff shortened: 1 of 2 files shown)


packages/oui-calendar/src/calendar.controller.js
```diff
@@ -100,6 +100,28 @@ export default class {
         this.initCalendarInstance();
     }
 
+    $onChanges ({ minDate, maxDate }) {
+        console.log(this.flatpickr);
```
frenautvh: please remove console.log

AxelPeter: Nice catch. That's the bad side of willing doing this fast :(

PR #453, opened 2020-01-07
fix(oui-datagrid): use lodash get to check properties

### LINKED ISSUE #448: oui-datagrid - unable to use getters for rows
### Description

When using a property defined with a getter as a column property, the column appears empty

### Steps to reproduce

1. Create a class 
2. Add a getter on that class
3. Create a datagrid and use the get as column property

```js
// model.js 
export default class Model {
  get property() {
    return 'awesome';
  }
}

// controller.js 
export default Controller {

 $onInit() {
   this.rows = [new Model()];
 }
}
```

```html
<!-- index.html -->
<oui-datagrid
  data-rows="$ctrl.rows">
  <oui-column
    data-title="'Column'"
   data-property="property"
 ></oui-column>
</oui-datagrid>
```

**Expected behavior**:

The property should be used and displayed in the column

**Actual behavior**:

The cell is empty 

**Frequency**:
Always

marie-j:
This allow to also use getters as properties

fixes #448

### DIFF
--- a/packages/oui-datagrid/src/util.js
+++ b/packages/oui-datagrid/src/util.js
@@ -1,11 +1,6 @@
+import get from "lodash/get";
+import isUndefined from "lodash/isUndefined";
+
 export function hasProperty (obj, prop = "") {
-    const props = prop.split(".");
-    let theObject = obj;
-    for (let i = 0; i < props.length; i++) {
-        if (!Object.prototype.hasOwnProperty.call(theObject, props[i])) {
-            return false;
-        }
-        theObject = theObject[props[i]] || {};
-    }
-    return true;
+    return !isUndefined(get(obj, prop));
 }


packages/oui-datagrid/src/util.js
```diff
@@ -1,11 +1,6 @@
...
-    const props = prop.split(".");
-    let theObject = obj;
-    for (let i = 0; i < props.length; i++) {
-        if (!Object.prototype.hasOwnProperty.call(theObject, props[i])) {
-            return false;
-        }
-        theObject = theObject[props[i]] || {};
-    }
-    return true;
+    return !isUndefined(get(obj, prop));
```
frenautvh: https://lodash.com/docs/4.17.15#has ?

marie-j: It doesn't have the same result because the checks made are not the same. `has` uses `Object.prototype.hasOwnProperty` which will result the same. 
Whereas `get` makes some checks by checking some `enumerable` property belongs to object (https://github.com/lodash/lodash/blob/master/.internal/isKey.js#L24)

marie-j: It doesn't have the same result because the checks made are not the same. `has` uses `Object.prototype.hasOwnProperty` which will result the same. 
Whereas `get` makes some checks by checking some `enumerable` property belongs to object (https://github.com/lodash/lodash/blob/master/.internal/isKey.js#L24)

FredericEspiau:
Good initiative

AxelPeter:
Don't forget to port it on v3 ;)

packages/oui-datagrid/src/util.js
frenautvh: @marie-j and what about https://lodash.com/docs/4.17.15#hasIn ? it looks the same : https://github.com/lodash/lodash/blob/master/hasIn.js

marie-j: @frenautvh did not see that one ! I'll try it out and update if it's the case

PR #455, opened 2020-01-09
Feat/collapsible loading

AxelPeter:
* Add `loading` attribute
* Add `group` attribute
* Update documentation
* Update unit tests

ref: UK-122

### DIFF
--- a/packages/oui-collapsible/src/collapsible.controller.js
+++ b/packages/oui-collapsible/src/collapsible.controller.js
@@ -1,9 +1,10 @@
 import { addDefaultParameter, removeHtmlTags } from "@ovh-ui/common/component-utils";
 
 export default class {
-    constructor ($attrs, $element, $scope, $timeout, $window) {
+    constructor ($attrs, $document, $element, $scope, $timeout, $window) {
         "ngInject";
         this.$attrs = $attrs;
+        this.$document = $document;
         this.$element = $element;
         this.$scope = $scope;
         this.$timeout = $timeout;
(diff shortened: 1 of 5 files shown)


packages/oui-collapsible/src/collapsible.html
```diff
@@ -1,11 +1,12 @@
 <button class="oui-button oui-button_full-width oui-collapsible__header"
     type="button"
-    ng-click="$ctrl.toggle()"
+    ng-click="!$ctrl.loading && $ctrl.toggle()"
```
FredericEspiau: The button is not disabled ?

packages/oui-collapsible/src/collapsible.controller.js
```diff
@@ -42,7 +43,22 @@ export default class {
             .on("resize", () => this.$scope.$apply());
     }
 
+
+    close () {
+        if (this.expanded) {
```
FredericEspiau: I feel like if this method is called and the component is not open, then we should send an error, instead of ignoring the programming error: https://en.wikipedia.org/wiki/Fail-fast

AxelPeter: This method is mainly there for the accordion mode and don't need to send error. 
All we want, is to have each item of the same group in the `closed` state before opening a new one. If there is one or more opened, we toggle it (and send the event callback). We don't want to get an error each time it's called if the item is already closed.

In this case, we just call it for each component of the same group without checking before if it's opened or not.

packages/oui-collapsible/src/collapsible.html
AxelPeter: No, we just disable the click during a loading. IMO, the disabled state should be something different. But both has the same effect (unclickable button) :)

PR #458, opened 2020-01-30
feat(oui-datagrid): refresh datagrid

radireddy:
add component support to refresh datagrid
MANAGER-4342

### Requirements

### DIFF
--- a/packages/oui-datagrid/src/datagrid.directive.js
+++ b/packages/oui-datagrid/src/datagrid.directive.js
@@ -19,7 +19,8 @@ export default () => ({
         onRowSelect: "&",
         onPageChange: "&",
         onCriteriaChanged: "&onCriteriaChange",
-        onSortChange: "&"
+        onSortChange: "&",
+        refreshable: "<?"
     },
     compile: elm => {
         // Transclude can't be used here otherwise transcluded
--- a/packages/oui-datagrid/src/datagrid.html
+++ b/packages/oui-datagrid/src/datagrid.html
@@ -7,6 +7,14 @@
     <oui-datagrid-extra-top class="oui-datagrid-topbar__slot"
         ng-if="$ctrl.hasExtraTopContent">
     </oui-datagrid-extra-top>
+    <button
+        ng-if="$ctrl.refreshable"
+        class="oui-datagrid-topbar__refresh oui-button oui-button_secondary oui-button_icon-only oui-button_small-width"
+        aria-label="::$ctrl.config.translations.ariaRefreshButtonLabel"
+        ng-click="$ctrl.refreshData(false, false, true, true);"
+    >
+        <span class="oui-icon oui-icon-refresh_extra-thin" aria-hidden="true"></span>
+    </button>
 </oui-criteria>
 <div class="oui-datagrid-responsive-container"
     ng-class="{
--- a/packages/oui-datagrid/README.md
+++ b/packages/oui-datagrid/README.md
@@ -596,6 +596,29 @@ You can use `row-loader`. It take the current row as argument and must return a 
 
 ### Refresh
 
+Datagrid has inbuilt refresh button. Clicking on it, calls row-loader and rows-loader. This will re-fetch the data. To enable this icon, set refreshable attribute on oui-datagrid.
+
+```html:preview
+<oui-datagrid
+  page-size="5"
+  id="localRefreshDatagrid"
+  rows="$ctrl.data"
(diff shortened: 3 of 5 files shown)


packages/oui-datagrid/src/datagrid.directive.js
```diff
@@ -19,7 +19,9 @@ export default () => ({
         onRowSelect: "&",
         onPageChange: "&",
         onCriteriaChanged: "&onCriteriaChange",
-        onSortChange: "&"
+        onSortChange: "&",
+        refreshable: "@?",
```
AxelPeter: Since it's a boolean, this should be `refreshable: "<?"`.
See for example the `customizable` attribute.

In the same way, we should support `<oui-datagrid refreshable></oui-datagrid>` as a `true` value. We have a helper addBooleanParameter for that: https://github.com/ovh-ux/ovh-ui-angular/blob/master/packages/oui-datagrid/src/datagrid.controller.js#L80

packages/oui-datagrid/src/datagrid.directive.js
```diff
@@ -19,7 +19,9 @@ export default () => ({
...
+        refreshIconAreaLabel: "@?"
```
AxelPeter: I think there is no need of an attribute for this. This should be only set in the provider, like you did: eg. https://github.com/ovh-ux/ovh-ui-kit/blob/master/packages/components/password/src/js/password.provider.js#L7

To keep a consistency with the others `aria` translations, it could be clearer to be renamed `ariaResfreshButtonLabel` or `ariaRefreshLabel` (see Password examples). What do you think ?

packages/oui-datagrid/src/datagrid.html
```diff
@@ -7,6 +7,14 @@
     <oui-datagrid-extra-top class="oui-datagrid-topbar__slot"
         ng-if="$ctrl.hasExtraTopContent">
     </oui-datagrid-extra-top>
+    <button
+        data-ng-if="$ctrl.refreshable"
```
AxelPeter: We don't prefix with `data-` in our templates :)

packages/oui-datagrid/src/datagrid.html
```diff
@@ -7,6 +7,14 @@
...
+        class="oui-datagrid-topbar__refresh oui-button oui-button_secondary oui-button_icon-only oui-button_small-width"
+        aria-label="::$ctrl.refreshIconAreaLabel || $ctrl.config.translations.refreshIconAreaLabel"
+        data-ng-click="$ctrl.refreshData(false, false, true, true);"
```
AxelPeter: Same as above

packages/oui-datagrid/src/datagrid.html
```diff
@@ -7,6 +7,14 @@
...
+    >
+        <span class="oui-icon oui-icon-refresh_extra-thin" aria-hidden="true"></span>
```
AxelPeter: We can remove `aria-hidden` since it's the only of the button

packages/oui-datagrid/src/datagrid.directive.js
radireddy: makes sense, we should maintain the consistency

packages/oui-datagrid/src/datagrid.html
radireddy: I think it should be hidden from screen readers as aria-label is provided in button and screen reader reads it. It does not need to know about the refresh icon. it's only for people who can see it.

packages/oui-datagrid/src/datagrid.html
```diff
@@ -7,6 +7,14 @@
     <oui-datagrid-extra-top class="oui-datagrid-topbar__slot"
         ng-if="$ctrl.hasExtraTopContent">
     </oui-datagrid-extra-top>
+    <button
+        ng-if="$ctrl.refreshable"
+        class="oui-datagrid-topbar__refresh oui-button oui-button_secondary oui-button_icon-only oui-button_small-width"
+        aria-label="::$ctrl.config.translations.ariaRefreshButtonLabel"
+        ng-click="$ctrl.refreshData(false, false, true, true);"
```
FredericEspiau: 😅

packages/oui-datagrid/src/datagrid.html
FredericEspiau: You could actually remove it because `aria-label` will overruler everything in the content of the HTML element 🙂

PR #460, opened 2020-02-14
feat(oui-select-picker): add transclusion for label

JeremyDec:
Signed-off-by: Jérémy De-Cesare <jeremy.de-cesare@corp.ovh.com>

Backport on V3 can be done after the acceptation of this changes.
Some changes from prettier has been done

### DIFF
--- a/packages/oui-select-picker/src/select-picker.html
+++ b/packages/oui-select-picker/src/select-picker.html
@@ -26,9 +26,10 @@
         </span>
 
         <!-- Label -->
-        <span class="oui-select-picker__label">
-            <span id="{{:: $ctrl.id }}Label"
-                ng-bind="$ctrl.label">
+        <span class="oui-select-picker__label"
+              id="{{:: $ctrl.id }}Label"
+              ng-transclude="labelSlot">
+            <span ng-bind="$ctrl.label">
             </span>
         </span>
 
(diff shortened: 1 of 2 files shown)


packages/oui-select-picker/src/select-picker.controller.js
```diff
@@ -37,23 +42,34 @@ export default class SelectPickerController {
...
+        return (
+            this.$transclude.isSlotFilled("descriptionSlot") || this.description
+        );
     }
 
     hasFooter () {
         return this.$transclude.isSlotFilled("footerSlot") || this.footer;
     }
 
+    hasLabel () {
```
JeremyDec: Only this has been added

packages/oui-select-picker/src/select-picker.html
```diff
@@ -7,76 +9,100 @@
...
+            />
+            <span ng-class=":: $ctrl.picture" ng-if="!$ctrl.isImgPath"> </span>
         </span>
 
         <!-- Label -->
-        <span class="oui-select-picker__label">
-            <span id="{{:: $ctrl.id }}Label"
-                ng-bind="$ctrl.label">
-            </span>
+        <span
```
JeremyDec: Only this part has changed

AxelPeter:
@JeremyDec can you revert the changes from prettier and only add the needed changes?

packages/oui-select-picker/src/select-picker.html
```diff
@@ -26,7 +26,9 @@
         </span>
 
         <!-- Label -->
-        <span class="oui-select-picker__label">
+        <span class="oui-select-picker__label"
+              id="{{:: $ctrl.id }}Label"
+              ng-transclude="labelSlot">
             <span id="{{:: $ctrl.id }}Label"
```
AxelPeter: Careful, you'll get a double ID here. You could remove the one on this `span` ;)

PR #464, opened 2020-04-29
feat(oui-navbar): add onClick attribute for brand component

frenautvh:
### Description of the Change

Add onClick attribute for the brand component inside the oui-navbar

### Benefits

Allow to add callback on brand logo click for example : tracking

### Applicable Issues

internal MANAGER-4921 tracking is required


packages/oui-navbar/src/brand/navbar-brand.html
```diff
@@ -1,7 +1,8 @@
 <a class="oui-navbar__brand"
     ng-attr-aria-label="{{::!!$ctrl.ariaLabel ? $ctrl.ariaLabel : null}}"
     ng-attr-title="{{::!!$ctrl.ariaLabel ? $ctrl.ariaLabel : null}}"
-    ng-href="{{::$ctrl.href}}">
+    ng-href="{{::$ctrl.href}}"
+    ng-click="$event.preventDefault(); $ctrl.onNavigateLink()">
```
AxelPeter, line 5: Why can't we have a simple `ng-click="$ctrl.onClick()"`, like we have here https://github.com/ovh-ux/ovh-ui-angular/blob/master/packages/oui-navbar/src/link/navbar-link.html#L13 ?

To avoid having functional part in the component, the `preventDefault` should be in the callback function in the Manager.

packages/oui-navbar/src/brand/navbar-brand.component.js
```diff
@@ -22,6 +25,16 @@ export default {
                     .removeAttr("aria-label")
             );
         }
+
+        onNavigateLink () {
+            if (angular.isFunction(this.onClick)) {
+                this.$q.when(this.onClick()).finally(() => {
```
AxelPeter, line 31: I think this part should be in the Manager side. It seems a bit complex for just adding a click function on a link :)

Normally, `ng-click="$ctrl.onClick()"` should be enough. We could even pass the event parameter if needed like  `ng-click="$ctrl.onClick($event)"`.

packages/oui-navbar/src/brand/navbar-brand.component.js
```diff
@@ -7,13 +7,16 @@ export default {
         iconAlt: "@?",
         iconClass: "@?",
         iconSrc: "@?",
-        href: "@?"
+        href: "@?",
+        onClick: "<?"
```
AxelPeter, line 11: For callback function, prefer using this.

```suggestion
        onClick: "&"
```

We don't put `&?` because the fallback is already done with AngularJS with `angular.noop()`. Or else we need to check before calling it :)

packages/oui-navbar/src/navbar.html
```diff
@@ -16,7 +16,8 @@
         aria-label="{{::$ctrl.brand.label}}"
         icon-alt="{{::$ctrl.brand.iconAlt}}"
         icon-class="{{::$ctrl.brand.iconClass}}"
-        icon-src="{{::$ctrl.brand.iconSrc}}">
+        icon-src="{{::$ctrl.brand.iconSrc}}"
+        on-click="::$ctrl.brand.onClick">
```
AxelPeter, line 20: I think we should keep the same property name as the other links: `click`. Or else we will have some surprises if we have in one hand `.click` and in the other `.onClick`.

https://github.com/ovh-ux/ovh-ui-angular/blob/master/packages/oui-navbar/src/navbar.html#L38

An idea could to support both, but needs to be updated on the others links. What do you think ?

frenautvh: yep done

frenautvh:
> Just need the documentation about this new attribute and we're all good :)

just updated the doc

